PR #47, opened 2015-09-18
Corrected text measurements, changed descent sign

npoltorapavlo:


cfry002:
Bob was working on a few tests we want to use to verify baseline before merging these changes.

PR #67, opened 2016-02-18
Added new px:scene.1-based support for RPC per getRPCContext spec

tjcarroll2:
px:scene.1-based support for RPC per getRPCContext spec.
rpc = scene.getRPCContext();
rpc.registerApp(appId);
rpc.registerFunction(functionName, filters);
rpc.execute(appId, functionName, args, callback);

### DIFF
--- a/examples/pxScene2d/src/jsbindings/rcvrcore/rpcContext.js
+++ b/examples/pxScene2d/src/jsbindings/rcvrcore/rpcContext.js
@@ -1,0 +1,98 @@
+/**
+ * Created by tjcarroll2
+ * on 2/6/16.
+ */
+
+function RPCContext(theScene) {
+  var scene = theScene;
+  var rpcController = null;
+  var sourceTargetName;
+  var appNameTargetRegistered = false;
+  var functions = {};
+  var onRPCCall;
+
+  this.registerApp = function(targetName) {
+    if( rpcController !== null && !appNameTargetRegistered ) {
+      appNameTargetRegistered = rpcController.registerApp(targetName, scene);
+      sourceTargetName = targetName;
+    } else {
+      console.error("RPC registerApp: No rpcController exists or target app is already registered for " + targetName + ", rpcController=" + rpcController);
+    }
+
+    return appNameTargetRegistered;
+  };
+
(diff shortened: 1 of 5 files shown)


examples/pxScene2d/src/jsbindings/rcvrcore/rpcContext.js
```diff
@@ -0,0 +1,98 @@
...
+
+    return appNameTargetRegistered;
+  };
+
+  this.onRPCCall = function(callback) {
+    onRPCCall = callback;
+  };
+
+  this.registerFunction = function(functionName, domainFiltersArray) {
+    if( rpcController !== null || !appNameTargetRegistered ) {
```
cfry002: Shouldn't this be (rpcController !== null && appNameTargetRegistered)?

examples/pxScene2d/src/jsbindings/rcvrcore/rpcContext.js
```diff
@@ -0,0 +1,98 @@
...
+
+function RPCContext(theScene) {
+  var scene = theScene;
+  var rpcController = null;
+  var sourceTargetName;
+  var appNameTargetRegistered = false;
+  var functions = {};
+  var onRPCCall;
+
+  this.registerApp = function(targetName) {
```
cfry002: I wonder if something like appNamespace or appId might be a better name for this var than "targetName".

examples/pxScene2d/src/jsbindings/rcvrcore/rpcContext.js
tjcarroll2: Good catch!  I don't know what I was thinking there.

examples/pxScene2d/src/jsbindings/rcvrcore/rpcContext.js
tjcarroll2: Bob called it appId on the Confluence page.  I think I like appNamespace better, since it implies it purpose.  I was originally looking at a general purpose RPC target that was more object based, but since we're taking the general onRPCCall approach this doesn't make sense anymore.

PR #78, opened 2016-04-07
build rtCore and rtRPC shared libraries

goruklu:
updated RPC makefile to build rtCore and rtRpc modules as shared
libraries.


mfiess:
Looks good

mfiess:
Merged to _temp_pxcore2_stage.

PR #82, opened 2016-04-20
Fixing a crash/memory corruption

vertexodessa:
When iterating with p++, in pxBuffer::fill, the assumption is to have 4-byte increment, which is not always true for "long" values

### DIFF
--- a/src/pxPixels.h
+++ b/src/pxPixels.h
@@ -6,6 +6,7 @@
 #define PX_PIXELS_H
 
 #include "pxCore.h"
+#include <stdint.h>
 
 struct pxPixel {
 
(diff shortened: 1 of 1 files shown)


src/pxPixels.h
```diff
@@ -53,7 +54,7 @@ struct pxPixel {
             unsigned char b: 8;
 #endif
         };
-        unsigned long u;
+        uint32_t u;
```
vertexodessa: unsigned long size is often 8 bytes on 64-but machine (I'm running 64-bit Linux Mint) 
http://stackoverflow.com/questions/7607502/sizeoflong-in-64-bit-c

johnrobinsn:
Thanks... nice catch.

JR

PR #83, opened 2016-04-22
fix bug in numeric_cast() function

goruklu:
In rtRpcConfig.cpp numeric_cast() function,
assert(l <= max) was failing on a 32bit processor (max needs to unsigned)


johnrobinsn:
Jake can you review this one way or the other?

gladish:
This was fixed in another change.

PR #88, opened 2016-05-11
Fix build with support for render hinting and delegation in westeros

emutavchi:


emutavchi:
already fixed by https://github.com/johnrobinsn/pxCore/pull/87

PR #95, opened 2016-05-26
Initial version of google test cases are added for RPC Component

rrao002c:
Initial version of google test cases are added for RPC component


rrao002c:
Initial version of google test cases are added for RPC Component. 
Note: To run google test cases, need to install and build google mock/test lib

PR #96, opened 2016-05-27
Decrease socket latency

vertexodessa:


examples/pxScene2d/src/rpc/rtRemoteStream.cpp
```diff
@@ -256,6 +260,10 @@ rtError
 rtRemoteStream::connectTo(sockaddr_storage const& endpoint)
 {
   m_fd = socket(endpoint.ss_family, SOCK_STREAM, 0);
+  uint32_t one = 1;
```
vertexodessa: just realized that it makes sense to move this code below m_fd check. will do..

PR #98, opened 2016-06-03
Avoid setting TCP_NODELAY option to sockets with non-TCP types

vertexodessa:
also, refactoring: moving the tcp option setting to utils module


johnrobinsn:
Jake can you review

PR #100, opened 2016-06-06
Added google test cases for Objectcache

rrao002c:
Added google test cases for Objectcache class


rrao002c:
Added google test cases for Objectcache class

PR #110, opened 2016-06-24
Added int64_t and uint64_t rtValue support

valbok:
After the commit it would be possible to call methods with int64 and uint64 params.


valbok:
After the commit it would be possible to call methods with int64 and uint64 params.

PR #114, opened 2016-07-07
Exclude westeros-stub impl when westeros_compositor is found in SYS_ROOT

vivekarumugam:
westeros-stub is always being built and have to preload system provided libwesteros_compositor.so


vivekarumugam:
Duplicate of https://github.com/johnrobinsn/pxCore/pull/112

PR #133, opened 2016-08-17
 temp refactor

madanagopalt:


aazamansari:
What happens when a wayland application does not use rtRemote for communication?

madanagopalt:
For non-rt remote app, it will run as usual.But no rtremote communications are made. This is tested with westeros_test from javascript.Thanks.

madanagopalt:
closing

PR #136, opened 2016-08-31
Generate static libs with "_s" suffix

vivekarumugam:
In addition to that, added cleanobjdir as dependency for pxscene.so, pxwayland.so & pxwayland_s.a. 

This is to remove previously built object files before building either pxscene/pxwayland as they both have different compilation switches ("-DENABLE_RT_NODE")

### DIFF
--- a/examples/pxScene2d/src/Makefile
+++ b/examples/pxScene2d/src/Makefile
@@ -8,8 +8,7 @@ deploy: pxscene.dmg
 else
 all: pxscene
 dfb: pxscene-dfb
-libs: librtCore.so libpxscene.so
-static-libs: librtCore.a libpxwayland.a
+libs: libpxscene.so
 endif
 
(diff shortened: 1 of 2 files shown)


mfiess:
I think the build scripts build librtCore.so and not lrtCore_s when building librtCore and libpxscene.so.  We should keep libpxscene.so using the .so version instead of the .a version for now.

vivekarumugam:
* Added librtRemote_s.a and librtCore_s.a as internal dependencies for libpxscene.so, libpxwayland.so and libpxwayland_s.a. In that way we need not to externally build those targets.

* Added separate rules to clean pxscene and pxwayland objects. This is done for the following reasons
  1) pxwayland and pxscene has different compilation parameters ("-DENABLE_RT_NODE")
  2) If we built pxwayland.so/pxwayland_s.a after building pxscene.so, previously built objects were taken for processing
  3) Cleaning entire obj/ directory leads to unnecessarily building rtCore_s.a again

PR #139, opened 2016-09-12
Memory leak changes with updated code for download request removal

madanagopalt:


examples/pxScene2d/src/jsbindings/rtWrapperUtils.cpp
```diff
@@ -111,36 +111,42 @@ void weakCallback_rt2v8(const WeakCallbackData<Object, rtIObject>& data)
...
-    // rtLogInfo("looking at:%d == %d", ref->CreationContextId, contextId);
+      ObjectReference* ref = begin->second;
+      rtLogInfo("looking at:%d == %d", ref->CreationContextId, contextId);
 
-    if (ref->CreationContextId == contextId)
-    {
-      ref->PersistentObject.ClearWeak();
-      ref->PersistentObject.Reset();
-      delete ref;
-      objectMap.erase(begin++);
```
mfiess: Can we do: begin = objectMap.erase(begin); 

instead of adding a new vector<iterator> refs?

PR #140, opened 2016-09-13
Fix for a crash problem: terminateClient makes use of the context which is destroyed by the call to WstCompositorDestroy.

GlebNovodranPE:
Changing the call order seems to help to eliminate the crash. Please review.


vertexodessa:
Could you please also create a pull to _temp_refactor_odessa branch? We're going to switch there soon

PR #144, opened 2016-09-20
Some makefile fixes

vertexodessa:


examples/pxScene2d/src/Makefile
```diff
@@ -286,6 +288,9 @@ PXSCENE_LIB_SEARCH=\
   -I$(NODEINC)/deps/debugger-agent/include \
   -I$(NODEINC)/deps/v8/include \
   -I$(NODEINC)/deps/cares/include \
+  -I$(EXTDIR)/ft/include \
```
vertexodessa: I was trying to fix libpxscene.so, but it's probably not needed, going to remove.

PR #149, opened 2016-09-23
Changes for fixing texture and pxImage memory leaks in pxscene

madanagopalt:


examples/pxScene2d/src/pxScene2d.cpp
```diff
@@ -1121,6 +1121,15 @@ rtDefineObject(pxRoot,pxObject);
 
 int gTag = 0;
 
+#ifdef ENABLE_RT_NODE
+rtNode script;
+#endif //ENABLE_RT_NODE
+
+static void pxScene2d::garbageCollect()
+{
+  script.garbageCollect();
```
mfiess: Please surround script.garbageCollect() with #ifdef ENABLE_RT_NODE.

examples/pxScene2d/src/pxScene2d.cpp
```diff
@@ -1121,6 +1121,15 @@ rtDefineObject(pxRoot,pxObject);
 
 int gTag = 0;
 
+#ifdef ENABLE_RT_NODE
+rtNode script;
+#endif //ENABLE_RT_NODE
+
+static void pxScene2d::garbageCollect()
```
mfiess: Is this API needed?  Can't we just call script.garbageCollect() directly?

examples/pxScene2d/src/pxScene2d.h
```diff
@@ -1260,15 +1264,13 @@ class pxScene2d: public rtObject, public pxIView
 
   rtMethodNoArgAndNoReturn("dispose",dispose);
 
+  static void garbageCollect();
```
mfiess: Please remove this API under pxScene2d.  We can access this by calling script.garbageCollect() directly.

PR #156, opened 2016-09-29
Build fix and unwiring pxwayland from libnode/pxscene

vertexodessa:
Two commits;
First one fixes typo which breaks the build in runtime
Second unwires pxWayland from libnode and pxscene

### DIFF
--- a/examples/pxScene2d/src/Makefile
+++ b/examples/pxScene2d/src/Makefile
@@ -278,7 +278,7 @@ PXWAYLAND_LIB_SRCS=\
            pxContextGL.cpp
 
 ifneq ($(DISABLE_WAYLAND),TRUE)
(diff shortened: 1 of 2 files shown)


examples/pxScene2d/external/libpng-1.6.12/libpng-config
```diff
@@ -17,7 +17,7 @@ exec_prefix="${prefix}"
 libdir="${exec_prefix}/lib"
 includedir="${prefix}/include/libpng16"
 libs="-lpng16"
-all_libs="-lpng16 -lz "
+all_libs="-lpng16 -lm -lz -lm "
```
vertexodessa: unintentional change, will remove

PR #161, opened 2016-10-04
Build both shared and static versions of the libraries

Gavriliuk:
Build both shared and static versions of libraries rtCore, rtRemote, pxwayland and pxscene simultaneously


examples/pxScene2d/src/Makefile
```diff
@@ -8,7 +8,7 @@ deploy: pxscene.dmg
 else
 all: pxscene
 dfb: pxscene-dfb
-libs: librtCore.so libpxscene.so
+libs: libpxscene.so libpxscene_s.a libpxwayland.so libpxwayland_s.a
```
mfiess: The libs target is built by the rng150 as well (which does not support pxWayland).  Please remove libpxwayland from this list and call it directly from the recipe when building to ensure builds that do not support wayland do not break.

Gavriliuk: ok, it's clear

PR #174, opened 2016-10-21
Flags added for enabling/disabling httpcache and fix for random hung …

madanagopalt:
…issue


examples/pxScene2d/src/Makefile
```diff
@@ -142,6 +142,10 @@ SEARCH=\
 #CXXFLAGS=-Wall -Wextra -g $(SEARCH) -DRT_PLATFORM_LINUX -D$(PX_PLATFORM) -DENABLE_RT_NODE
```
mfiess: I don't believe any Makefile changes are needed at this time.  Let's omit all Makefile changes.

examples/pxScene2d/src/pxFileCache.h
```diff
@@ -1,3 +1,4 @@
+#ifdef ENABLE_HTTP_CACHE
```
mfiess: Let's remove this #if condition.  I think it's best if we compile everything by default.  The ENABLE_HTTP_CACHE should just enable this capability, but it should be compiled by default (to ensure others don't break the API contract accidentally).

examples/pxScene2d/src/pxHttpCache.cpp
```diff
@@ -1,3 +1,4 @@
+#ifdef ENABLE_HTTP_CACHE
```
mfiess: Let's remove this #if condition.  I think it's best if we compile everything by default.  The ENABLE_HTTP_CACHE should just enable this capability, but it should be compiled by default (to ensure others don't break the API contract accidentally).

examples/pxScene2d/src/pxHttpCache.h
```diff
@@ -1,3 +1,4 @@
+#ifdef ENABLE_HTTP_CACHE
```
mfiess: Let's remove this #if condition.  I think it's best if we compile everything by default.  The ENABLE_HTTP_CACHE should just enable this capability, but it should be compiled by default (to ensure others don't break the API contract accidentally).

examples/pxScene2d/src/pxFileCache.cpp
```diff
@@ -1,3 +1,4 @@
+#ifdef ENABLE_HTTP_CACHE
```
mfiess: Let's remove this #if condition.  I think it's best if we compile everything by default.  The ENABLE_HTTP_CACHE should just enable this capability, but it should be compiled by default (to ensure others don't break the API contract accidentally).

PR #179, opened 2016-11-14
Add a callback to get notification on remote termination/closure

vivekarumugam:


mfiess:
Looks good

examples/pxScene2d/src/rpc/rtRemoteServer.cpp
```diff
@@ -523,6 +537,10 @@ rtRemoteServer::findObject(std::string const& objectId, rtObjectRef& obj, uint32
         // sure that this really matters much
         std::unique_lock<std::mutex> lock(m_mutex);
         m_object_map.insert(ClientMap::value_type(endpointName, client));
+
+        clientDisconnectedCB.func = cb;
+        clientDisconnectedCB.data = cbdata;
+        m_disconnected_callback_map.insert(ClientDisconnectedCBMap::value_type(client.get(), &clientDisconnectedCB));
```
aazamansari: clientDisconnectedCB has to be client specific and not server specific. Still the issue related to callback setting is not fixed as per my previous comments.

clientDisconnectedCB has to be a map and each client should be notified using its separate callback. With current code it won't work if we connect to multiple clients .ie., if we use rtRemoteLocateObject() to locate different remote objects and . 

Callback information shouid be stored inside rtRemoteClient structure and not in server.

examples/pxScene2d/src/rpc/rtRemoteServer.h
```diff
@@ -84,7 +87,14 @@ class rtRemoteServer
     bool              owner_removed;
   };
 
+  struct ClientDisconnectedCB
+  {
+      clientDisconnectedCallback func;
```
aazamansari: store just these 2 values in rtRemoteClient. on remote state change you can give callback from there which would be specific to client.

PR #180, opened 2016-11-15
Shutdown crash patch

vertexodessa:


examples/pxScene2d/src/rpc/rtRemoteClient.h
```diff
@@ -74,7 +76,6 @@ class rtRemoteClient: public std::enable_shared_from_this<rtRemoteClient>
     { return reinterpret_cast<rtRemoteClient *>(argp)->onStreamStateChanged(stream, state); }
 
   rtError onIncomingMessage(rtRemoteMessagePtr const& msg);
-  rtError onStreamStateChanged(std::shared_ptr<rtRemoteStream> const& stream, rtRemoteStream::State state);
```
vertexodessa: this is not required; occasionally moved the line.

examples/pxScene2d/src/rpc/rtRemoteClient.cpp
```diff
@@ -143,6 +142,7 @@ rtRemoteClient::onStreamStateChanged(std::shared_ptr<rtRemoteStream> const& /*st
 rtError
 rtRemoteClient::setStateChangedHandler(StateChangedHandler handler, void* argp)
```
vertexodessa: this can be called from the destructor, where we're holding mutex, so had to change the mutex type to recursive

PR #193, opened 2016-12-09
Call rtRemoteLocate on main thread

mgopalakirshnan:
Please review this change regarding WKIT-745

### DIFF
--- a/examples/pxScene2d/src/pxWayland.cpp
+++ b/examples/pxScene2d/src/pxWayland.cpp
@@ -21,11 +21,6 @@ using namespace std;
 
 extern pxContext context;
 
-#define MAX_FIND_REMOTE_TIMEOUT_IN_MS 5000
-#define FIND_REMOTE_ATTEMPT_TIMEOUT_IN_MS 100
-#define TEST_REMOTE_OBJECT_NAME "waylandClient123" //TODO - update
-
-
 pxWayland::pxWayland()
   :
     mRefCount(0),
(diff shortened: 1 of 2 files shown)


examples/pxScene2d/src/pxWayland.cpp
```diff
@@ -569,45 +524,29 @@ rtError pxWayland::startRemoteObjectLocator()
 rtError pxWayland::connectToRemoteObject()
 {
   rtError errorCode = RT_FAIL;
-#ifdef ENABLE_PX_WAYLAND_RPC
-  int findTime = 0;
 
-  while (findTime < MAX_FIND_REMOTE_TIMEOUT_IN_MS && mClientPID != -1)
+#ifdef ENABLE_PX_WAYLAND_RPC
+  rtLogInfo("Attempting to find remote object %s", mRemoteObjectName.cString());
+  errorCode = rtRemoteLocateObject(mRemoteObjectName.cString(), mRemoteObject, 1000);
```
vertexodessa: I believe I've seen some long child app startups when search was taking over 1 second, it's safer to set it to some higher value (maybe 5 sec)
Thanks!

mgopalakirshnan: In Wayland-Plugin, following Macros are used to keep on locating remote object for 5 seconds(5000 MS) at the interval of 100MS.
**_#define MAX_FIND_REMOTE_TIMEOUT_IN_MS 5000
#define FIND_REMOTE_ATTEMPT_TIMEOUT_IN_MS 100_**

PR #202, opened 2016-12-22
support exporting variables from standalone module

madanagopalt:


madanagopalt:
closing this as there is duplicate pull request in 238 which address some more comments on this change

PR #207, opened 2017-01-17
Node inspector support for pxscene from intelliJ

madanagopalt:
Please review the changes and provide your comments

### DIFF
--- a/examples/pxScene2d/src/pxScene.cpp
+++ b/examples/pxScene2d/src/pxScene.cpp
@@ -15,7 +15,11 @@
 #include "pxUtil.h"
 
 #ifdef RUNINMAIN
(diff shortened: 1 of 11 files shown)


examples/pxScene2d/src/pxScene.cpp
```diff
@@ -15,7 +15,11 @@
 #include "pxUtil.h"
 
 #ifdef RUNINMAIN
+#ifdef ENABLE_DEBUG_MODE
+extern rtNode* script;
```
mfiess: Let's either make script always a rtNode* or make it always an object (no matter if ENABLE_DEBUG_MODE is enabled or not)

madanagopalt:
closing this as changes are pushed from other pull request

PR #257, opened 2017-03-09
Fix for retrying socket connection for all entries after dns lookup

binuinbaraj:


binuinbaraj:
This may not be needed

PR #258, opened 2017-03-09
Pass ip Mode when doing a dnslookup 

binuinbaraj:


binuinbaraj:
Will re-commit with comments added

PR #271, opened 2017-03-19
Improved visual quality for scaled text and bugs

johnrobinsn:
See textscale.js example.
- Separated glyph texture caching from glyph metric caching.
- Glyph textures are no longer one for one with a glyph's pixelSize.  A given pixelSize glyph may use the texture from a larger pixelSize glyph.  This reduces overall memory and provides more pixel data for sampling.
- Glyph rendering is done using the exact metrics for the specific pixelSize as reported by free type.  The selected glyph texture (which may be for a larger pixelSize) is scaled (down) to fit inside the glyph bounding box for the specificed pixelSize.
- Glyph texture selection now takes into account the necessary LOD (Level of Detail) needed based on sx and sy scale factors.  Greatly improving visual quality when scaling text.

### DIFF
--- a/examples/pxScene2d/src/pxFont.cpp
+++ b/examples/pxScene2d/src/pxFont.cpp
@@ -305,6 +305,8 @@ const GlyphCacheEntry* pxFont::getGlyph(uint32_t codePoint)
       entry->advancedoty = g->advance.y;
       entry->vertAdvance = g->metrics.vertAdvance; // !CLF: Why vertAdvance? SHould only be valid for vert layout of text.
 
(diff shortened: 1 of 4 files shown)


johnrobinsn:
Connie can you review this and run through all of the text measurement regression tests and merge it if all things look good.

conniefry:
Functionally, the tests look good so far; but, I'm seeing a significant degradation in performance in some tests.

johnrobinsn:
Which tests are you seeing a performance hit?

johnrobinsn:
Fixed a bug where glyph cache wasn't being populated.

PR #274, opened 2017-03-20
developer tool changes (debug log test)

madanagopalt:


examples/pxScene2d/src/shell.js
```diff
@@ -145,6 +145,14 @@ if (false)
         childScene.url = homeURL;
         e.stopPropagation();
       }
+      else
+      if(code == keys.D)  // ctrl-alt-shft-d
+      {
+        if (process.env.PX_DUMP_MEMUSAGE && (process.env.PX_DUMP_MEMUSAGE == "1"))
```
mfiess: Let's remove the environment variable check.  This Easter egg should log the metrics even if the environment variable is not set.

PR #280, opened 2017-03-28
Don't do hole punching when drawing without FBO

emutavchi:
When rendering to default FBO the post rendering hole punching will clear everything, for example when there is a westerossink playing fullscreen.


johnrobinsn:
Remind me to talk try o you about this

PR #284, opened 2017-03-31
Changes for integrating breakpad for Linux platform

binuinbaraj:


examples/pxScene2d/src/Makefile
```diff
@@ -202,6 +203,12 @@ LDEXT += -lX11 $(LDLIBJPEGTURBO)
 endif
 endif
 
+ifeq ($(UNAME_S),Linux)
```
mfiess: Let's change this check to ifeq ($(PX_PLATFORM),PX_PLATFORM_GLUT) to ensure these libraries are not included for non-glut platforms.

PR #286, opened 2017-03-31
Windows project and installer

jmgasper:
Note that the Windows exe still has some bugs / crashing issues, but this should be a good start.


conniefry:
I'm going to merge this pull request, but I want to document a few known issues with it that need to be fixed in subsequent merges:

- In the latest revision of package instructions, it says: "Download https://go.microsoft.com/fwlink/?LinkId=746571 and put in folder `pxCore\pxScene2d`" . But there is no 'pxCore\pxScene2d' folder.  Should it be 'pxCore\pxCore.vsbuild\pxScene2d' ?

- Also, thanks for renaming the setup exe to pxscene-setup.exe; package instructions should be updated, too. 

- Auto update says, "Should pxScene Example App automatically check for updates? ..."  It should say, "Should pxscene automatically check for updates?..."  (Notice all lowercase pxscene as well as removal of "Example").

PR #294, opened 2017-04-06
fix for crash on app termination in mac

madanagopalt:


madanagopalt:
addressed review comments as part of other pull request

PR #311, opened 2017-04-14
 temp api error checks

FitzerIRL:
Added error checking of API args


FitzerIRL:
Messy lineage... I'll do-over from a Fresh FORK ....

PR #315, opened 2017-04-18
wayland apps configuration support

madanagopalt:


madanagopalt:
addressed review comments and raised a new request

PR #323, opened 2017-04-20
memory leak due to delegate for mac

madanagopalt:


src/mac/pxWindowNative.mm
```diff
@@ -1067,6 +1067,8 @@ -(void)drawRect:(NSRect)dirtyRect {
 pxError pxWindow::term()
 {
   NSWindow* window = (NSWindow*)mWindow;
+  WinDelegate* delegate = (WinDelegate*)mDelegate;
+  [delegate release];
```
mfiess: We should set the window's delegate to nil before releasing mDelegate ([window setDelegate:nil];).  We should also set mDelegate to nil after this.

PR #333, opened 2017-04-26
introducing animation api returning animation object

madanagopalt:
With this, below way of animation is introduced:

var animateObj = obj.animate();
animateObj.done.then();

Animation can be cancelled as below:
animateObj.cancel - true;

Below are the properties that can be queried:
done -> promise to indicate completion of animation
type -> type of animation
interp -> interpolators of animation
status -> status of animation IDLE,INPROGRESS,ENDED,CANCELLED
provduration -> provisioned duration of animation by end user
provcount -> provisioned count of animation by end user
cancel -> status of animation, is it cancelled or not
props -> properties added for animation like x,y etc
details -> details of status of animation of every property.

### DIFF
--- a/examples/pxScene2d/src/Makefile
+++ b/examples/pxScene2d/src/Makefile
@@ -105,7 +105,8 @@ PX_SRCS_FULL=\
   pxImage.cpp \
   pxImage9.cpp \
   pxImageA.cpp \
-  pxArchive.cpp 
+  pxArchive.cpp \
+  pxAnimate.cpp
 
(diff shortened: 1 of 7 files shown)


conniefry:
I would like to see some tests added to CI runs for this - either testRunner or C++  or both.  Thanks!

bobseamon:
+Narendra

Connie – Are these new properties official?  It might be a little early, but if so, it would be great to have them in the specs.

Thanks!

Bob

PR #343, opened 2017-04-28
change to make sure scene without parent gets deleted

madanagopalt:


madanagopalt:
this is for reviewing the approach

PR #348, opened 2017-05-04
running javascript files under unittests for linux

madanagopalt:


madanagopalt:
No, the variables g_argv,g_argv are serving to share the command line arguments needed for debugging to pxscene from unitests code.

Newly introduced variables gargc,gargv are serving to pass the complete input to pxscene for unittests.Since, we compile unittests with debug mode to validate debug related code with unittests, we may not remove it.

PR #357, opened 2017-05-09
increase code coverage for few components

madanagopalt:
please review


mfiess:
The code changes look good.  Can you please confirm that the images you are adding to the repository do not have any copyright or license concerns?  Can we download these images instead of adding them to the repository?

PR #358, opened 2017-05-10
 windows port

jmgasper:
This should fix the issues with:

* Linux build not working
* Naming of the output to consistently use "pxscene"


examples/pxScene2d/src/pxScene.cpp
```diff
@@ -195,27 +195,12 @@ class sceneWindow : public pxWindow, public pxIViewContainer
 #endif 
    // pxScene.cpp:104:12: warning: deleting object of abstract class type ‘pxIView’ which has non-virtual destructor will cause undefined behaviour [-Wdelete-non-virtual-dtor]
 
-#ifdef RUNINMAIN
```
conniefry: Why did you remove this?

conniefry: And the other gc related code, below?

PR #361, opened 2017-05-12
XRE2-10: Added support for ServiceManager

sborushevsky:


examples/pxScene2d/src/Makefile
```diff
@@ -414,6 +419,10 @@ PXSCENE_EXTRA_DEFINES=-DPNG_APNG_SUPPORTED
 endif
 endif
 
+ifeq ($(HAVE_RT_REMOTE),1)
+EXTRA_COMPILER_FLAGS:=$(EXTRA_COMPILER_FLAGS) -DUSE_RT_REMOTE 
+LDFLAGS_FULL:=$(LDFLAGS_FULL) -L../../../remote -lrtRemote
```
mfiess: This will work when building the static version of the libs (libpxscene.a), but it won't be included when building the shared library version (libpxscene.so).  We build the shared libraries when developing to allow for faster work environments.  Can you please update to ensure librtRemote is linked when building the shared library version as well?

sborushevsky: Enabled Service Manager support for libpxscene.so.
Actually I've enabled it to 3 targets only: pxscene, libpxscene.so and libpxscene.a.

examples/pxScene2d/src/Makefile
```diff
@@ -15,12 +15,15 @@ libs-glut: libpxscene-glut
...
 endif
 
+
 HAVE_WESTEROS := $(shell test -f ../external/westeros/Makefile.ubuntu && echo 1)
 
 ifneq ($(HAVE_WESTEROS),1)
 HAVE_WESTEROS := $(shell pkg-config --exists westeros-compositor && echo 1)
 endif
 
+HAVE_RT_REMOTE := $(shell test -f ../../../remote/librtRemote.so -o -f ../../../remote/librtRemote_s.a && echo 1)
```
mfiess: We should probably use a different means of determining whether or not to build with rtRemote support.  These libraries will be built for most platforms, but are not included in all.  Perhaps, we can use an environment variable here called SERVICE_MANAGER.  If we want to build with the service manager then we run: make SERVICE_MANAGER=1.  We can then test to see if this environment variable is set inside the Makefile.  If set then we will compile in pxServiceManager.cpp plus link in the rtRemote libs.

examples/pxScene2d/src/Makefile
```diff
@@ -106,6 +109,7 @@ PX_SRCS_FULL=\
   pxImage9.cpp \
   pxImageA.cpp \
   pxArchive.cpp \
+  pxServiceManager.cpp \
```
mfiess: pxServiceManager.cpp should be conditionally added based on the SERVICE_MANAGER environment variable.  Please refer to the comment above.

examples/pxScene2d/src/Makefile
```diff
@@ -264,6 +268,7 @@ PXSCENE_LIB_SRCS=\
   pxImageA.cpp \
   pxArchive.cpp \
   pxFont.cpp \
+  pxServiceManager.cpp \
```
mfiess: pxServiceManager.cpp should be conditionally added based on the SERVICE_MANAGER environment variable.  Please refer to the comment above.

examples/pxScene2d/src/pxScene2d.cpp
```diff
@@ -40,6 +40,9 @@
 #include "pxText.h"
 #include "pxTextBox.h"
 #include "pxImage.h"
+#ifdef USE_RT_REMOTE
```
mfiess: Please change USE_RT_REMOTE to PX_SERVICE_MANAGER.

examples/pxScene2d/src/pxScene2d.cpp
```diff
@@ -1532,6 +1535,10 @@ rtError pxScene2d::create(rtObjectRef p, rtObjectRef& o)
     e = createTextBox(p,o);
   else if (!strcmp("image",t.cString()))
     e = createImage(p,o);
+#ifdef USE_RT_REMOTE
```
mfiess: Please change USE_RT_REMOTE to PX_SERVICE_MANAGER.

examples/pxScene2d/src/pxScene2d.cpp
```diff
@@ -1611,6 +1618,14 @@ rtError pxScene2d::createImage(rtObjectRef p, rtObjectRef& o)
   return RT_OK;
 }
 
+#ifdef USE_RT_REMOTE
```
mfiess: Please change USE_RT_REMOTE to PX_SERVICE_MANAGER.

examples/pxScene2d/src/pxServiceManager.cpp
```diff
@@ -0,0 +1,29 @@
+#ifdef USE_RT_REMOTE
```
mfiess: We no longer need to surround the whole source file with a #ifdef since pxServiceManager will only be compiled in conditionally based on values passed to the Makefile.  Please refer to the comments above.

examples/pxScene2d/src/pxServiceManager.h
```diff
@@ -0,0 +1,21 @@
+#ifndef PX_SERVICEMANAGER_H
+#define PX_SERVICEMANAGER_H
+
+#ifdef USE_RT_REMOTE
```
mfiess: We no longer need to surround the whole source file with a #ifdef since pxServiceManager will only be compiled in conditionally based on values passed to the Makefile.  Please refer to the comments above.

PR #368, opened 2017-05-18
automate fix for leaks from scenes without parent

kannanmadanagopalt:
This is to address scenario, where scene is created without parent.

### DIFF
--- a/examples/pxScene2d/src/pxScene2d.h
+++ b/examples/pxScene2d/src/pxScene2d.h
@@ -986,12 +986,7 @@ public:
     return c;
   }
 
-  void dispose()
-  {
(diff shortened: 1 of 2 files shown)


examples/pxScene2d/src/pxScene2d.h
```diff
@@ -1479,6 +1491,7 @@ class pxScene2d: public rtObject, public pxIView
   int32_t mPointerHotSpotY;
   #endif
   bool mPointerHidden;
+  std::vector<pxSceneContainer*> mSceneConts;
```
mfiess: Please rename mSceneConts to mSceneContainers to allow for easier readability.

PR #371, opened 2017-05-19
automate fix for leaks from scenes without parent

madanagopalt:


examples/pxScene2d/src/pxScene2d.h
```diff
@@ -1479,6 +1491,7 @@ class pxScene2d: public rtObject, public pxIView
   int32_t mPointerHotSpotY;
   #endif
   bool mPointerHidden;
+  std::vector<pxSceneContainer*> mSceneContainers;
```
mfiess: Should this be pxSceneContainerRef instead of pxSceneContainer*?

examples/pxScene2d/src/pxScene2d.cpp
```diff
@@ -411,36 +411,39 @@ void pxObject::createNewPromise()
...
+    for(vector<rtRef<pxObject> >::iterator it = mChildren.begin(); it != mChildren.end(); ++it)
+    {
+      (*it)->dispose();
+      (*it)->mParent = NULL;  // setParent mutates the mChildren collection
+    } 
+    mChildren.clear();
+    deleteSnapshot(mSnapshotRef); 
+    deleteSnapshot(mClipSnapshotRef);
+    deleteSnapshot(mDrawableSnapshotForMask);
+    deleteSnapshot(mMaskSnapshot);
```
FitzerIRL: It seems like `deleteSnapshot()` is not really delete... given the reference is disposed of outside `deleteSnapshot()` with an explicit NULL assignment.   Perhaps renaming it to `clearSnapshot()` might be clearer.   Just a thought.

PR #375, opened 2017-05-22
Temp warnings as errors

FitzerIRL:
Changes to support WARNINGS as ERRORS 

Test builds on Ubuntu and Mac look good.


examples/pxScene2d/src/pxScene2d.cpp
```diff
@@ -1681,7 +1681,12 @@ rtError pxScene2d::logDebugMetrics()
 {
   script.garbageCollect();
   rtLogInfo("pxobjectcount is [%d]",pxObjectCount);
+
+#ifdef PX_PLATFORM_MAC
+  rtLogInfo("texture memory usage is [%lld]",context.currentTextureMemoryUsageInBytes());
```
mfiess: Let's change %lld to PRId64.  This is the desired define to use to print int64_t, and it will eliminate the need for a platform #ifdef here.

examples/pxScene2d/src/pxScene.cpp
```diff
@@ -198,7 +200,12 @@ class sceneWindow : public pxWindow, public pxIViewContainer
     if (gDumpMemUsage)
     {
       rtLogInfo("pxobjectcount is [%d]",pxObjectCount);
+
+#ifdef PX_PLATFORM_MAC
+      rtLogInfo("texture memory usage is [%lld]",context.currentTextureMemoryUsageInBytes());
```
mfiess: Let's change %lld to PRId64.  This is the desired define to use to print int64_t, and it will eliminate the need for a platform #ifdef here.

FitzerIRL: FYI - *PRId64* does NOT seem to work on the RNG150.

PR #379, opened 2017-05-24
fix for cancel status for animation with type oscillate

madanagopalt:


examples/pxScene2d/src/pxAnimate.h
```diff
@@ -24,6 +24,29 @@
 #include "pxConstants.h"
 class pxObject;
 struct animation;
+
+static rtString mapStatus(uint32_t status)
```
mfiess: Can we please move this function implementation into the cpp file?  We're looking to keep header files limited to definitions only with their implementations included in cpp files.

examples/pxScene2d/src/pxAnimate.h
```diff
@@ -24,6 +24,29 @@
...
+{
```
FitzerIRL: This `if / else` ladder code should be a `switch` statement.  Please change it.

examples/pxScene2d/src/pxAnimate.h
```diff
@@ -85,28 +108,25 @@ class pxAnimate: public rtObject
...
-
     rtObjectRef mProps;
     rtObjectRef mCurrDetails;
     uint32_t mInterp;
     pxConstantsAnimation::animationOptions mType;
     double mProvisionedDuration;
     int32_t mProvisionedCount;
     bool mCancelled;
-    rtString mStatus;
+    uint32_t mStatus;
```
FitzerIRL: `mStatus` is clearly an `enum` ...  which is a good thing ... strongly typed.  Please use that `enum` or a `typedef` of it instead of `uint32_t` wherever you're using or passing it about

examples/pxScene2d/src/pxAnimate.h
FitzerIRL: `switch(v)
{
  case pxConstantsAnimation::STATUS_IDLE:  return "IDLE";
  case pxConstantsAnimation::STATUS_CANCELLED:  return "CANCELLED";
  case pxConstantsAnimation::STATUS_ENDED:  return "ENDED";
  default: return "UNKNOWN";
}
`

PR #384, opened 2017-05-25
Update waylandregistry mechanism

jeffwannamaker:
In order to facilitate Westeros CTS testing, the waylandregistry mechanism is updated to behave as it does now if a registry match is found, but if no match is found, to use the cmd string as-is.

Signed-off-by: Jeff Wannamaker <jeff_wannamaker@cable.comcast.com>


jeffwannamaker:
Abandoning this pull request.  Will rework and resubmit.

PR #386, opened 2017-05-26
Fix for DFB surface LEAK in pxTextureOffscreen

FitzerIRL:


examples/pxScene2d/src/pxContextDFB.cpp
```diff
@@ -901,18 +962,28 @@ void decodeTextureData(void* data)
   if (data != NULL)
   {
     DecodeImageData* imageData = (DecodeImageData*)data;
+    
+    if(imageData == NULL)
```
mfiess: Is this check needed?  It should be covered under (data != NULL).

PR #394, opened 2017-05-31
enabling curl reuse handles

madanagopalt:


src/rtFileDownloader.h
```diff
@@ -147,7 +147,7 @@ class rtFileDownloader
     void (*mDefaultCallbackFunction)(rtFileDownloadRequest*);
     std::vector<rtFileDownloadHandle> mDownloadHandles;
     bool mReuseDownloadHandles;
-
+    std::string mCaCertFile;
```
mfiess: Please use rtString instead of std::string.

PR #453, opened 2017-06-22
Fix issues reported by Coverity scan

conniefry:


examples/pxScene2d/src/pxContextGL.cpp
```diff
@@ -2374,7 +2376,7 @@ int64_t pxContext::currentTextureMemoryUsageInBytes()
 
 int64_t pxContext::textureMemoryOverflow(pxTextureRef texture)
 {
-  int64_t textureSize = (texture->width()*texture->height()*4);
+  int64_t textureSize = (int64_t)(texture->width()*texture->height()*4);
```
mfiess: I think we should change this to the following to fix the coverity report:

int64_t textureSize = ((int64_t)(texture->width())*((int64_t))(texture->height())*4);

PR #455, opened 2017-06-23
Backtrace osx stack

madanagopalt:


madanagopalt:
opened for testing purpose

PR #467, opened 2017-07-03
fix for leak of pxobjects

madanagopalt:


examples/pxScene2d/src/pxScene2d.h
```diff
@@ -1499,7 +1499,7 @@ class pxScene2d: public rtObject, public pxIView
   int32_t mPointerHotSpotY;
   #endif
   bool mPointerHidden;
-  std::vector<pxSceneContainerRef> mSceneContainers;
+  std::vector<rtObjectRef> mInnerObjs;
```
mfiess: Please use full names as opposed to abbreviations when creating variables and methods.  Please change mInnerObjs to mInnerObjects.

examples/pxScene2d/src/pxScene2d.h
```diff
@@ -1451,7 +1451,7 @@ class pxScene2d: public rtObject, public pxIView
     return e;
   }
 
-  void sceneContainerDisposed(pxSceneContainerRef ref);
+  void innerObjDisposed(rtObjectRef ref);
```
mfiess: Please use full names as opposed to abbreviations when creating variables and methods.  Please change innerObjDisposed to innerObjectDisposed.

PR #470, opened 2017-07-10
fix for pxobjects leak

madanagopalt:


src/jsbindings/rtWrapperUtils.cpp
```diff
@@ -242,7 +240,14 @@ rtWrapperSceneUpdateEnter();
   pthread_mutex_lock(&sObjectMapMutex);
 #endif
   ObjectReferenceMap::iterator i = objectMap.find(from.getPtr());
-  assert(i == objectMap.end());
+  if (i != objectMap.end())
+  {
+    if (!(i->second->PersistentObject.IsNearDeath()))
+    {
+      assert(false);
```
mfiess: Let's try to better handle the error rather than throwing an assert() as this will crash the application.

PR #478, opened 2017-07-13
Enable compiling with llvm toolchain

andrewbenton:
Rapidjson's RAPIDJSON_HAS_CXX11_RVALUE_REFS define statement was bad.
Rapidjson has fixed this in later versions.

When not compiled on a mac platform, the `#pragma gcc` statement caused
the clang build to fail.  This statement is now ignored under llvm's
toolchaein

### DIFF
--- a/remote/rapidjson/rapidjson.h
+++ b/remote/rapidjson/rapidjson.h
@@ -498,8 +498,12 @@ RAPIDJSON_NAMESPACE_END
 
 #ifndef RAPIDJSON_HAS_CXX11_RVALUE_REFS
 #if defined(__clang__)
-#define RAPIDJSON_HAS_CXX11_RVALUE_REFS __has_feature(cxx_rvalue_references) && \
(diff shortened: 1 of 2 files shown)


remote/rapidjson/rapidjson.h
```diff
@@ -498,8 +498,12 @@ RAPIDJSON_NAMESPACE_END
 
 #ifndef RAPIDJSON_HAS_CXX11_RVALUE_REFS
 #if defined(__clang__)
-#define RAPIDJSON_HAS_CXX11_RVALUE_REFS __has_feature(cxx_rvalue_references) && \
+#if __has_feature(cxx_rvalue_references) && \
```
mfiess: rapidjson is a third party library that pxscene uses.  Please surround third party code changes with
/* MODIFIED CODE BEGIN */
/* MODIFIED CODE END */

PR #481, opened 2017-07-14
Skip unit tests and code coverage for release builds

conniefry:


conniefry:
Merging this so we can get a release build done even though codecov is complaining.

PR #521, opened 2017-08-02
enable leak checks for mac

madanagopalt:


ci/monitor.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/sh
+count=0
+while [ "$count" -ne 3600 ]; do
```
mfiess: Let's change not equals to less than just in case we change the interval in the future and the while is not updated.

madanagopalt: addressed

PR #537, opened 2017-08-08
Updated gitignore list

binuinbaraj:


andrewbenton:
Can this listing be more generalized with wildcards?

PR #544, opened 2017-08-10
  Added Data download progress callback.

ssureshmsc:
Added data download progress callback, so user can get the callback while download in progress.

### DIFF
--- a/src/rtFileDownloader.cpp
+++ b/src/rtFileDownloader.cpp
@@ -53,6 +53,7 @@ struct MemoryStruct
         , headerBuffer(NULL)
         , contentsSize(0)
         , contentsBuffer(NULL)
+        , downloadRequest(NULL)
     {
         headerBuffer = (char*)malloc(1);
         contentsBuffer = (char*)malloc(1);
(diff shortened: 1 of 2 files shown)


src/rtFileDownloader.h
```diff
@@ -79,6 +81,8 @@ class rtFileDownloadRequest
 #ifdef ENABLE_HTTP_CACHE
   void setCacheEnabled(bool val);
   bool cacheEnabled();
+  void setDataCache(bool val);
+  bool dataCached();
```
mfiess: Can you please rename setDataCache() to setDataIsCached() and dataCache() to isDataCached()?

src/rtFileDownloader.cpp
```diff
@@ -76,6 +76,7 @@ struct MemoryStruct
   char* headerBuffer;
   size_t contentsSize;
   char* contentsBuffer;
+  rtFileDownloadRequest *rtFileDownloadReq;
```
mfiess: Let's rename rtFileDownloadReq to downloadRequest to keep with the naming convention found in the other sources.  Also, can you please initialize this variable to NULL in the constructor?

ssureshmsc:
Fixed the review comments, please review.

PR #547, opened 2017-08-11
Fix for a crash on a client destruction

vertexodessa:
Client has to have possibility to unregister it's disconnect callback.
Otherwise when a client is destroyed manually (but not through Stream->disconnect),
callback might fire when a client is actually dead, leading to a crash.


remote/rtRemoteServer.cpp
```diff
@@ -579,6 +579,46 @@ rtRemoteServer::findObject(std::string const& objectId, rtObjectRef& obj, uint32
 }
 
 rtError
+rtRemoteServer::unregisterDisconnectedCallback( clientDisconnectedCallback cb, void *cbdata )
```
vertexodessa: I've just realized that we don't need the callback pointer to do the job, just a data pointer is required. Should I fix this?

gladish:
Do we need a function to remove the queue ready handler also?

vertexodessa:
@gladish hi Jake, I think 
rtRemoteRegisterQueueReadyHandler ( nullptr, nullptr )
will have the same effect as deregistering the queue ready handler

PR #552, opened 2017-08-14
changes for generating cache online

madanagopalt:


madanagopalt:
To generate cache on need, below command need to be executed from desktop:

python generate_cache.py --repo=pxCore --user=<user name> --apitoken=<token generated for repo>

To generate token for a repo in LINUX desktop, below steps are needed:
a,sudo apt install ruby ruby-dev
b,sudo gem install travis
c,travis login --org (here we need to provide the github username and password of required repository for which build needs to be  triggered).
d,travis token --org

PR #564, opened 2017-08-21
disabling local http and https access for remote files

madanagopalt:


FitzerIRL:
NOTE:  "::1" or in longhand, "0:0:0:0:0:0:0:1" is the loopback address in the IPv6 protocol. 

Not sure if the can be checked for in the same way... but might be worth trying.

Could be an attack vector for hacking.

PR #579, opened 2017-09-01
Added ability to set thread-specific error

gladish:
For APIs that don't naturally return an error, I added the capability to set a thread-specific rtError. This is similar in function to errno() or SetLastError().

Use in cases like this.

rtObject* obj = someClass.someFunction();
if (!obj)
   printf("error: %s", rtStrError(rtGetLastError()));

since this (contrived) API doesn't return the actual error, the implementer of that API can use rtErrorSetLastError() to stash the current error in thread-local and allow the caller to fetch it later if desired.

### DIFF
--- a/remote/Makefile
+++ b/remote/Makefile
@@ -67,6 +67,9 @@ ifeq ($(PROFILE), 1)
   CXXFLAGS += -pg
 endif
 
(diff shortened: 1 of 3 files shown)


mfiess:
I spoke with Jake, and he said that he believes that this change was brought in as a part of another PR.  Closing this PR.

PR #583, opened 2017-09-05
Remove unnecessary logging

andrewbenton:
Logging from both c++ and javascript was flooding the console and making
the output difficult to use.  This change removes many of the log
messages that were at fault, particularly rtLogWarn calls that remarked
about normal instantiation.

### DIFF
--- a/examples/pxScene2d/src/pxScene2d.cpp
+++ b/examples/pxScene2d/src/pxScene2d.cpp
@@ -2967,7 +2967,6 @@ void pxScriptView::runScript()
 // escape url end
 
   #ifdef ENABLE_RT_NODE
-  rtLogWarn("pxScriptView::pxScriptView is just now creating a context for mUrl=%s\n",mUrl.cString());
   mCtx = script.createContext();
 
   if (mCtx)
@@ -2988,7 +2987,6 @@ void pxScriptView::runScript()
     char buffer[MAX_URL_SIZE + 50];
     memset(buffer, 0, sizeof(buffer));
     snprintf(buffer, sizeof(buffer), "loadUrl(\"%s\");", mUrl.cString());
-    rtLogWarn("pxScriptView::runScript calling runScript with %s\n",mUrl.cString());
 #ifdef WIN32 // process \\ to /
 		unsigned int bufferLen = strlen(buffer);
 		char * newBuffer = (char*)malloc(sizeof(char)*(bufferLen + 1));
--- a/examples/pxScene2d/src/shell.js
+++ b/examples/pxScene2d/src/shell.js
@@ -74,8 +74,6 @@ if (false)
 	  var code  = e.keyCode;
     var flags = e.flags;
 
-    console.log("SHELL: onPreKeyDown:", code, " key: ", keys.name(code), ", ", flags);
-
     if( keys.is_CTRL_ALT( flags ) )
     {
       if(code == keys.Y)  // ctrl-alt-y
@@ -155,11 +153,8 @@ if (false)
 
   scene.root.on("onPreKeyUp", function(e)
   {
-    console.log("in onPreKeyUp", e.keyCode, e.flags);
 	  var code  = e.keyCode;
     var flags = e.flags;
-
-    console.log("onKeyUp:", code, ", ", flags);
 
     // eat the ones we handle here
          if (code == keys.Y && keys.is_CTRL_ALT( flags ) )       e.stopPropagation(); // ctrl-alt-y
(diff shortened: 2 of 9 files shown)


examples/pxScene2d/src/pxScene2d.cpp
```diff
@@ -2967,7 +2967,6 @@ void pxScriptView::runScript()
 // escape url end
 
   #ifdef ENABLE_RT_NODE
-  rtLogWarn("pxScriptView::pxScriptView is just now creating a context for mUrl=%s\n",mUrl.cString());
```
mfiess: Can you please convert this to rtLogInfo() instead of removing?

examples/pxScene2d/src/pxScene2d.cpp
```diff
@@ -2988,7 +2987,6 @@ void pxScriptView::runScript()
     char buffer[MAX_URL_SIZE + 50];
     memset(buffer, 0, sizeof(buffer));
     snprintf(buffer, sizeof(buffer), "loadUrl(\"%s\");", mUrl.cString());
-    rtLogWarn("pxScriptView::runScript calling runScript with %s\n",mUrl.cString());
```
mfiess: Can you please convert to rtLogInfo() instead of removing?

examples/pxScene2d/src/shell.js
```diff
@@ -74,8 +74,6 @@ if (false)
 	  var code  = e.keyCode;
     var flags = e.flags;
 
-    console.log("SHELL: onPreKeyDown:", code, " key: ", keys.name(code), ", ", flags);
```
mfiess: For the time being, let's leave the key prints in place as they are useful for debugging.

examples/pxScene2d/src/shell.js
```diff
@@ -155,12 +153,9 @@ if (false)
 
   scene.root.on("onPreKeyUp", function(e)
   {
-    console.log("in onPreKeyUp", e.keyCode, e.flags);
 	  var code  = e.keyCode;
```
mfiess: For the time being, let's leave the key prints in place as they are useful for debugging.

examples/pxScene2d/src/shell.js
```diff
@@ -155,12 +153,9 @@ if (false)
...
     var flags = e.flags;
 
-    console.log("onKeyUp:", code, ", ", flags);
```
mfiess: For the time being, let's leave the key prints in place as they are useful for debugging.

examples/pxScene2d/src/shell.js
```diff
@@ -175,7 +170,6 @@ if (false)
     scene.root.on("onKeyDown", function(e)
     {
       var code = e.keyCode; var flags = e.flags;
-      console.log("onKeyDown shell:", code, ", ", flags);
```
mfiess: For the time being, let's leave the key prints in place as they are useful for debugging.

examples/pxScene2d/src/shell.js
```diff
@@ -199,12 +193,9 @@ if (false)
 
   scene.root.on("onPreChar", function(e)
   {
-    console.log("in onchar");
 	  var c = e.charCode;
-    console.log("onChar:", c);
 	  // TODO eating some "undesired" chars for now... need to redo this
```
mfiess: For the time being, let's leave the key prints in place as they are useful for debugging.

examples/pxScene2d/src/shell.js
```diff
@@ -199,12 +193,9 @@ if (false)
 
   scene.root.on("onPreChar", function(e)
   {
-    console.log("in onchar");
```
mfiess: This log is a good one to remove.

examples/pxScene2d/src/shell.js
```diff
@@ -199,12 +193,9 @@ if (false)
...
 	  var c = e.charCode;
-    console.log("onChar:", c);
 	  // TODO eating some "undesired" chars for now... need to redo this
     if (c<32) {
-      console.log("stop onChar");
```
mfiess: For the time being, let's leave the key prints in place as they are useful for debugging.

examples/pxScene2d/src/shell.js
andrewbenton: Do we have a way to conditionally enable this?  They might be useful for debugging for developers of pxscene, but not so much for people developing applications on pxscene.

conniefry: Let's compromise and check the environment variable to know whether or not to log these messages:
- Use process.env.RT_LOG_LEVEL: Valid values are "debug", "info" , "warn" , "error", "fatal".  Only log if value is debug or info.
- Please add a consistent prefix for these messages while you're here; something like "SHELL:"; that will help people to grep these out when they wish. 
- Long term we should look at adding a logging API wrapper so we don't need a check for log level sprinkled throughout.

PR #587, opened 2017-09-08
fix a leak suppression with compiler optimizations

madanagopalt:
sometimes, compiler optimizations is adding few strings to the symbol names, which cause failure in leak suppression.


ci/testRunner_memcheck_osx.js
```diff
@@ -15,7 +15,7 @@ var testUrls = ["http://pxscene.org/examples/px-reference/gallery/fancy.js",
                 ];
 
 //added different timeout for different pages, as travis server is bit slower and so getting more timeout errors for complex js pages
-var testTimeouts = [5000, 5000, 20000, 420000];
+var testTimeouts = [5000, 5000, 20000, 600000];
```
FitzerIRL: A comment about units would be good here... 600000 Seconds ? or Milliseconds ?

More comments and whitespace help clarity

conniefry:
If the leak suppression is still necessary, please submit new PR. The other files in this PR have changed or have been deleted since this PR was created.

PR #590, opened 2017-09-13
updated build instructions for rtRemote

mfiess:


examples/pxScene2d/README.md
```diff
@@ -150,7 +150,34 @@ file:///home/username/directory/filename.js
     pxscene2dtests.exe
     ~~~~
 
+## Building rtRemote (Linux only)
```
conniefry: For clarity, I think this should say 'Building rtRemote only' and maybe clarify that you only need to do this if you need rtRemote outside of the pxscene binary....

PR #604, opened 2017-09-22
HTTP access control (CORS) for pxArchives

npoltorapavlo:


src/rtFileDownloader.cpp
```diff
@@ -655,6 +655,7 @@ bool rtFileDownloader::downloadFromNetwork(rtFileDownloadRequest* downloadReques
     if (false == headerOnly)
     {
       downloadRequest->setDownloadedData(chunk.contentsBuffer, chunk.contentsSize);
+      checkAccessControlHeaders(downloadRequest);
```
mfiess: Can we add a #ifdef around the checkAccessControlHeaders() call called #ifdef ENABLE_ACCESS_CONTROL_CHECK?  Then please add an option to CMakeLists.txt (called PXCORE_ACCESS_CONTROL_CHECK) that will enable this define when the option is turned on.  By default, this option should be turned on.  However, we would like the ability to quickly disable if needed.

PR #608, opened 2017-09-28
SVG Rounded Rectangles

FitzerIRL:
Adding support for SVG Rounded Rectangles ... some other optimizations

### DIFF
--- a/examples/pxScene2d/src/rcvrcore/scene.1.js
+++ b/examples/pxScene2d/src/rcvrcore/scene.1.js
@@ -59,7 +59,19 @@ function Scene() {
 
     if(params.hasOwnProperty("d") && params.t === "path")
     {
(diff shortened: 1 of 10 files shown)


examples/pxScene2d/src/rcvrcore/scene.1.js
```diff
@@ -59,7 +59,19 @@ function Scene() {
...
+        if(params.d.match(/rect/i) )
+        {
+          params.d = params.d.replace(/rect/gi, "RECT");
+          
+          // normalize the path
+          params.d = params.d.replace(/,/g," ")
+          .replace(/-/g," -")
+          .replace(/ +/g," ");
+          
+           console.log(" >>> Found RECT: [" + params.d + "] ");
```
mfiess: Let's remove this console.log() and only log for debugging.

PR #620, opened 2017-10-13
Fix expiration data in cacheHeader

conniefry:


mfiess:
The expiration date will be changed to programmable mock data when we remove the need for using a local web server.

PR #621, opened 2017-10-14
src/rtObjectMacros.h: Type address to uintptr_t for comparision in macro

kraj:
This is an output pointer which could be NULL but hard for compiler to
detect that in a macro

Ideally, macros in c++ are not encouraged, they should be converted into
static inlines, however if we want to keep using macros compilers can
warn about address comparison

error: the compiler can assume that the address of 'r' will always evaluate to 'true' [-Werror=address]

Signed-off-by: Khem Raj <raj.khem@gmail.com>

### DIFF
--- a/src/rtObjectMacros.h
+++ b/src/rtObjectMacros.h
@@ -184,13 +184,13 @@ int class::rtPropertyCount = sizeof(class::rtPropertyEntries)/sizeof(rtPropertyE
 // Internally used thunk macros
 // thunks for functions returning a value
 #define rtThunkNoArgAndReturn(method, returntype) \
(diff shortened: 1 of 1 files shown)


conniefry:
FYI:  I just restarted the Travis build (again) now that PR 620 has been merged.  It was at least partially responsible for failures in travis linux builds for this PR.

mdrost:
IMHO this pull lacks consistency. For example why macro rtThunk1ArgAndReturn is unchanged?

kraj:
@mdrost while you are right this is minimal fix that we needed to get going with newer compiler
It’s possible that the rtThunk1ArgAndReturn is unused and thus never compiled in any relevant source

PR #634, opened 2017-10-25
added alpha to image blend

BunioFH:
from now on image opacity will be taken into account
when drawing.


johnrobinsn:
I'm not sure if I understand why this is needed or desired  can i get more detail on why we need this...

conniefry:
Bunio, could you send a fail case for a scenario that requires this change?

PR #635, opened 2017-10-25
Bunio 9part

BunioFH:


conniefry:
What is the specific use case for this change?  We would want to ensure backward compatibility of this change with current authoring, as well.

conniefry:
Closing this PR in favor of PR #733 that implements image9Border for the functionality needed.

PR #637, opened 2017-10-25
Bunio even framedrop

BunioFH:


conniefry:
Bunio, could you please have a look at the changes made in commit d1e71675d41e36013446fe909a5c6a23bb0372fa on October 20th and see if they address your concerns and make this PR unnecessary?

BunioFH:
no they don't. the change you mentioned assumes that load will contain itself within a single frame (1) and (2) it still uses roughly 32ms per frame instead of even pacing causing a noticeable slip every second.

src/wayland_egl/pxWindowNative.cpp
```diff
@@ -540,6 +548,15 @@ void pxWindowNative::runEventLoopOnce()
 }
 
 
+static uint64_t getMicroseconds() {
```
FitzerIRL: **pxTimerNative.cpp**  has an implementation of  `pxMicroseconds()`

conniefry:
Closing in favor of pull request #671

PR #650, opened 2017-11-03
update vm and port for build logs uploads

conniefry:


conniefry:
Merging this change so that the old VM used for logs and artifacts can be shutdown.

PR #665, opened 2017-11-10
Modified strcpy to strncpy in rtFileDownloadRequest::setHTTPError

ssureshmsc:


mfiess:
This is being tracked in 667

PR #672, opened 2017-11-13
fix build warning (issue 668)

mfiess:


examples/pxScene2d/external/westeros-stub/westeros-compositor.h
```diff
@@ -59,7 +59,7 @@ typedef enum _WstHints
 typedef void (*WstTerminatedCallback)( WstCompositor *ctx, void *userData );
 typedef void (*WstDispatchCallback)( WstCompositor *ctx, void *userData );
 typedef void (*WstInvalidateSceneCallback)( WstCompositor *ctx, void *userData );
-typedef void (*WstDecodeHandlerCallback)( WstCompositor *ctx, void *userData, uint32_t decoderHandle);
+typedef void (*WstDecodeHandlerCallback)( WstCompositor *ctx, void *userData, void *decoderHandle);
```
dwrobel: This change seems to create discrepancy between official Westeros [WstDecodeHandlerCallback()](https://github.com/rdkcmf/westeros/blob/master/westeros-compositor.h#L59) declaration  - which will result in difficult to debug crashes on 64-bit platforms, if someone will use for compilation westeros-compositor.h header for pxScene and use westeros .so from the platform.

mfiess: This change shouldn't be merged until the official westeros declaration is updated.

mfiess:
Fixed as a part of 710.

PR #678, opened 2017-11-15
fix make 'clean' target

dwrobel:
Add missing deletion of rtRemoteConfigBuilder.cpp auto-generated file.


mfiess:
Fix looks good.  Retriggering validation build.

PR #680, opened 2017-11-15
fix shader compilation errors (issue 679)

dwrobel:
Patch removes "precision mediump float;" from ifdef-ed code
and attempts to compile source code with and without above
instruction (as a fallback).


mfiess:
This issue will be fixed by pull request 690.  A shader check will determine if it's OpenGL ES or not.  If the firmware is OpenGL ES then precision will be set.  Otherwise, it is not set.

PR #686, opened 2017-11-17
add missing 'uv' library

dwrobel:
Fixes the follwoing linking failure:
  /usr/bin/ld: ../../../build/glut/libpxCore.a(rtNode.cpp.o): undefined reference to symbol 'uv_default_loop'
  //usr/lib64/libuv.so.1: error adding symbols: DSO missing from command line
  collect2: error: ld returned 1 exit status


mfiess:
This has introduced a linking error for Ubuntu.  Please take a look

dwrobel:
This is better addressed by #700 .

PR #693, opened 2017-11-21
disable few signals handles in node to avoid process exit

madanagopalt:


mfiess:
This change is no longer needed.  This fix is included in pull request 698.

PR #700, opened 2017-11-22
changes to make it work with system libraries

dwrobel:

PR #701, opened 2017-11-22
fix include directory for v8

dwrobel:
1. It would make easier to use source files to compile with system Node.js.
2. Currently NODEINC already contains '${NODEDIR}/deps/v8/include' which
   allow us to include "v8.h" directly instead of "include/v8.h".


dwrobel:
This is required by https://github.com/pxscene/pxCore/pull/700

PR #721, opened 2017-12-07
changes for removing timers data on timeout

madanagopalt:


conniefry:
Do you have any tests to add to CI for verification?

PR #737, opened 2017-12-13
options to set pxscene resolution on startup

mfiess:


mfiess:
atoi() should work good in this case.  The code provides error checking to make sure correct values are set.  I agree that strtol() should be used in cases where we would want to handle overflow, etc.  However, the numbers passed here for the resolution should be small (e.g. 1280x720).  Plus there are error checks to ensure desired values are received when parsed.  In this case, atoi() should work good (along with the error checking that is in place).

mfiess:
I will update this code to use strtol() instead of atoi().

PR #742, opened 2017-12-16
Fix for missing ucontext -> ucontext_t

andrewbenton:
In modern versions of glibc, ucontext is defined exclusively as
ucontext_t.  This minor fix allows the build to work again.  I don't
know which version of glibc this starts with, but it effects glibc 2.26
and later at least and wasn't needed for glibc 2.24


mfiess:
Can you please retest with the latest version?  This should now be patched.

PR #748, opened 2017-12-18
fix for memory increase on page reloads

madanagopalt:


examples/pxScene2d/src/rcvrcore/XModule.js
```diff
@@ -168,12 +195,30 @@ XModule.prototype._importModule = function(requiredModuleSet, readyCallBack, fai
       } else {
         for (var k = 0; k < _this.moduleNameList.length; ++k) {
           ///_this.appSandbox[pathToNameMap[exports[k][1]]] = exports[k][0];
-          _this.moduleData[_this.moduleNameList[k]] = exports[k][0];
-          exportsArr[k] = exports[k][0];
+          //_this.moduleData[_this.moduleNameList[k]] = exports[k][0];
```
mfiess: Please add a comment why this is commented out and no longer needed for our future reference when we look at the code.

PR #761, opened 2018-01-02
Fix for pxObjectCount leak

arun-govindan:


arun-govindan:
That option was tried in OnCloseRequest function, but it was not resolving the issue.

PR #764, opened 2018-01-03
avoid using cached texture with dirty text object

madanagopalt:


examples/pxScene2d/src/pxText.cpp
```diff
@@ -174,7 +174,7 @@ void pxText::draw() {
   if( getFontResource() != NULL && getFontResource()->isFontLoaded())
   {
     // TODO not very intelligent given scaling
-    if (msx == 1.0 && msy == 1.0 && mCached.getPtr() && mCached->getTexture().getPtr())
+    if (!mDirty && msx == 1.0 && msy == 1.0 && mCached.getPtr() && mCached->getTexture().getPtr())
```
mfiess: Just want to confirm...  Will mDirty only be true for one frame?  After that will mCached contain the updated text fbo?

mfiess:
Madan confirmed that mDirty will only be dirty one frame.

PR #766, opened 2018-01-04
fix compilation and runtime errors with duktape

madanagopalt:


examples/pxScene2d/external/dukluv/src/main.c
```diff
@@ -69,7 +71,7 @@ static duk_ret_t duv_loadfile(duk_context *ctx) {
   if (fd) uv_fs_close(&loop, &req, fd, NULL);
   uv_fs_req_cleanup(&req);
   
-  duk_error(ctx, DUK_ERR_ERROR, "%s: %s: %s", uv_err_name(req.result), uv_strerror(req.result), path2.c_str());
+  duk_error(ctx, DUK_ERR_ERROR, "%s: %s: %s", uv_err_name(req.result), uv_strerror(req.result), path2);
 }
 
 struct duv_list {
```
mfiess: We should make external changes as a patch file instead of modifying directly.

PR #767, opened 2018-01-04
Ci optimization

arun-govindan:


arun-govindan:
Created new pull request as resolving conflicts is not allowed for this PR

PR #768, opened 2018-01-04
fix for screenshot's base64 string NULL-termination

npoltorapavlo:


mfiess:
Nikita, do we need to make the changes to pxScene2d.cpp?  In the mac, we need to add the NULL terminator when setting the encoded_data to ensure there isn't a crash during screen capture.

npoltorapavlo:
NULL terminator is now added by line "rtString base64str(d, l);" for all platforms.

PR #784, opened 2018-01-12
duktape support

mfiess:


examples/pxScene2d/src/CMakeLists.txt
```diff
@@ -85,13 +85,15 @@ if (APPLE)
     set(PLATFORM_LIBRARIES pxCore rtCore_s pthread png16 jpeg freetype curl z)
 
     if (SUPPORT_DUKTAPE)
+        add_definitions(-DRTSCRIPT_SUPPORT_NODE)
```
dwrobel: Is that ok? shouldn't it be -DRTSCRIPT_SUPPORT_DUKTAPE ?

src/wayland_egl/pxWindowNative.cpp
```diff
@@ -571,13 +571,15 @@ void pxWindowNative::runEventLoop()
 
     rtLogInfo("pxcore framerate: %d", framerate);
 
-    uint64_t* offsets = new  uint64_t[ framerate ];
-    for( int i = 0; i < framerate; ++i )
-        offsets[ i ] = (i*1000000+framerate-1)/framerate;
-
-    int frameNo = 1;
-    double wakeUpBase = pxMicroseconds();
-    int count = 0, lastCount = -1;
```
dwrobel: This breaks the compilation for BUILD_PXSCENE_WAYLAND_EGL:
`pxCore/src/wayland_egl/pxWindowNative.cpp:585:9: error: ‘count’ was not declared in this scope`

PR #800, opened 2018-01-19
External upload

arun-govindan:
Updates done:
1. Upload and reuse the built external
2. --without-docs flag added for upgrade cmake command.
3. Excluded ccache on mac os


arun-govindan:
At present travis performance is better and some optimisation points are also done to reduce the execution time.
If required this PR might be considered in future.

Closing this PR and corresponding ticket [XRE2-1194](https://ccp.sys.comcast.net/browse/XRE2-1194)

PR #810, opened 2018-01-25
Changes for enabling duktape from Makefile and fixing space related issues

madanagopaltcomcast:


andrewbenton:
Have you accidentally added all changed files into your commit?  All of the new files look like a copy of old files, a patch, a few make files, and some images across 34 files.  That doesn't seem to fit with changing a few space constraints.

PR #811, opened 2018-01-25
Provided access to application to use the FilePointer to read the data in a smaller chunk.

ssureshmsc:
The existing code  allocates buffer for the full file and returning to application for use.
 If the cached file is bigger file say 30 MB, it allocates for 30 MB and store the data for use.
 Now added the provision to application to read it however it wants. Also there is no change in existing file reading machanics unless application sets setReadCachedDataInChunks.

### DIFF
--- a/src/rtFileDownloader.cpp
+++ b/src/rtFileDownloader.cpp
@@ -153,7 +153,7 @@ rtFileDownloadRequest::rtFileDownloadRequest(const char* imageUrl, void* callbac
     mDownloadedData(0), mDownloadedDataSize(), mDownloadStatusCode(0) ,mCallbackData(callbackData),
     mCallbackFunctionMutex(), mHeaderData(0), mHeaderDataSize(0), mHeaderOnly(false), mDownloadHandleExpiresTime(-2)
 #ifdef ENABLE_HTTP_CACHE
-    , mCacheEnabled(true), mIsDataInCache(false)
+    , mCacheEnabled(true), mIsDataInCache(false), mReadCacheDataInChunks(false)
 #endif
     , mIsProgressMeterSwitchOff(false), mHTTPFailOnError(false), mDefaultTimeout(false)
 {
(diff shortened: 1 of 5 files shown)


src/rtFileDownloader.cpp
```diff
@@ -366,6 +366,29 @@ bool rtFileDownloadRequest::isDataCached()
...
 
+void rtFileDownloadRequest::setReadCachedDataInChunks(bool val)
+{
+  mReadCacheDataInChunks = val;
+}
+
+bool rtFileDownloadRequest::readCachedDataInChunksEnabled()
+{
+  return mReadCacheDataInChunks;
+}
```
mfiess: Let's rename setReadCachedDataInChunks() and readCachedDataInChunksEnabled() APIs to setDeferCacheRead() and deferCacheRead().

src/rtFileDownloader.h
```diff
@@ -83,6 +83,9 @@ class rtFileDownloadRequest
   bool cacheEnabled();
   void setDataIsCached(bool val);
   bool isDataCached();
+  void setReadCachedDataInChunks(bool val);
+  bool readCachedDataInChunksEnabled();
```
mfiess: Let's rename setReadCachedDataInChunks() and readCachedDataInChunksEnabled() APIs to setDeferCacheRead() and deferCacheRead().

src/rtHttpCache.cpp
```diff
@@ -222,7 +222,7 @@ rtData& rtHttpCacheData::contentsData()
   return mData;
 }
 
-rtError rtHttpCacheData::data(rtData& data)
+rtError rtHttpCacheData::data(rtData& data, bool readCachedDataInChunks)
```
mfiess: Let's not modify the data() API or call.  Instead, let's create an API on rtHttpCacheData that will setup it up to be read lazily.

tests/pxScene2d/test_imagecache.cpp
```diff
@@ -518,7 +518,7 @@ class rtHttpCacheTest : public testing::Test, public commonTestFns
       const char* cacheData = "This is image data";
```
mfiess: Please create new unit tests in test_imagecache.cpp to test the new APIs.  Adding code coverage for the new APIs will ensure our code coverage does not drop.

src/rtFileDownloader.cpp
```diff
@@ -783,7 +806,7 @@ bool rtFileDownloader::checkAndDownloadFromCache(rtFileDownloadRequest* download
   rtData data;
   if ((NULL != rtFileCache::instance()) && (RT_OK == rtFileCache::instance()->httpCacheData(downloadRequest->fileUrl(),cachedData)))
   {
-    err = cachedData.data(data);
+    err = cachedData.data(data, downloadRequest->readCachedDataInChunksEnabled());
```
mfiess: We should avoid this data() call if deferCacheRead() is true.  Instead, please call an API on rtHttpCacheData that will setup rtHttpCache to be read lazily.

PR #816, opened 2018-01-29
enable font memory saving strategy

madanagopaltcomcast:


mfiess:
This issue will be addressed with font atlas.

PR #844, opened 2018-02-05
duktape support for OSX and linux

arun-govindan:


arun-govindan:
Closing this PR, Since a new PR #875 is raised with a single build support

PR #846, opened 2018-02-06
Reverting decoderHandle passing from Pxcore. 

nambirajang:
Since Helio player is abandoned, We are reverting Helio CC supporting code changes.


nambirajang:
Raised new pull request #848.

PR #851, opened 2018-02-08
Unified type and template Real type-width use

FitzerIRL:


FitzerIRL:
Redundant - nanosvg supersedes this.

PR #852, opened 2018-02-09
resolve 'duktape merge hacks', reinstate CORS/permissions for duktape

npoltorapavlo:


examples/pxScene2d/src/rcvrcore/AppSceneContext.js
```diff
@@ -443,97 +301,113 @@ AppSceneContext.prototype.runScriptInNewVMContext = function (packageUri, module
...
+      }
+      log.message(4, "vm.runInNewContext done");
 
-      if (process._debugWaitConnect) {
+      if (!isDuk && process._debugWaitConnect) {
         // Set breakpoint on module start
         if (process.env.BREAK_ON_SCRIPTSTART != 1)
           delete process._debugWaitConnect;
         const Debug = vm.runInDebugContext('Debug');
         Debug.setBreakPoint(moduleFunc, 0, 0);
```
mfiess: It looks like moduleFunc is no longer set.

npoltorapavlo: moduleFunc is not needed now - I invoke it in place, see "(px, xModule, fname, this.basePackageUri)"

mfiess: It looks like moduleFunc is still being used in line 409 (in the new code).

npoltorapavlo: ok, restored moduleFunc

mfiess: The reference to moduleFunc on line 410 `(Debug.setBreakPoint(moduleFunc, 0, 0); ` is not defined in its scope.  Unfortunately, the variable for moduleFunc goes out of scope after the else statement above.  The line `Debug.setBreakPoint(moduleFunc, 0, 0);` will result in an undefined variable of moduleFunc.

mfiess: The use of var allows the scope for moduleFunc to be valid.

PR #860, opened 2018-02-12
Correct work and shutdown rtRemoteServer even if pipe was not opened

Gavriliuk:
If pipe was not opened in constructor rtRemote still works.
But it tries to read the unopened pipe and can't send a shutdown signal to the listener thread.

This change fixes usage of the pipe and offers an alternative way to shutdown the listener thread.


Gavriliuk:
We need not to send any special signal to wake up the server thread.
rtRemoteEnvironment sets m_running to false before stopping Server.
Checking of this value will stop the thread in case of unopened pipe.

Moreover, we can use for this a special new field in class rtRemoteServer.
This will be done in the following patch.

PR #862, opened 2018-02-13
duktape debugger changes [Ready for review]

madanagopalt:


src/CMakeLists.txt
```diff
@@ -34,6 +34,7 @@ set(BUILD_RTCORE_SHARED_LIBRARY 1)
 
 if (APPLE)
     message("Building pxcore for mac")
+    option(SUPPORT_DUKTAPE_DEBUGGER "SUPPORT_DUKTAPE_DEBUGGER" ON)
```
dwrobel: Will it be an option to consider reuse DISABLE_DEBUG_MODE which already exists and enables/disables inspector/debugger support for Node instead of creating yet another option.

The problem with new option is that it doesn't check at all if the SUPPORT_DUKTAPE is ON.

Additionally, SUPPORT_DUKTAPE_DEBUGGER doesn't work seamlessly with all backends (if I'm not mistaken it will work on Linux only for glut but not e.g. for Wayland).

madanagopalt: Thanks. Yes, I made similar option we have for node debug enablement for duktape.Currently, it is verified with linux,osx and windows.

src/CMakeLists.txt
```diff
@@ -53,6 +53,9 @@ if (APPLE)
             mac/pxClipboardNative.mm mac/window.mm
             unix/rtMutexNative.cpp unix/rtThreadPoolNative.cpp
             rtFileCache.cpp rtHttpCache.cpp)
+     	if (SUPPORT_DUKTAPE)
+          set(PXCORE_FILES ${PXCORE_FILES} rtScriptDuk/debugger/duk_debugger_interface_unix.c)
```
mfiess: Let's create two subdirectories under rtScriptDuk/debugger/.  One should be called win the other should be called unix (following the same pattern as platform specific files in pxCore/src).  Let's have both files named the same (duk_debugger_interface.c) and put the unix version under rtScriptDuk/debugger/unix/ and the windows version under rtScriptDuk/debugger/win/.

src/CMakeLists.txt
```diff
@@ -165,6 +174,9 @@ elseif(WIN32)
     set(PXCORE_FILES win/pxBufferNative.cpp win/pxOffscreenNative.cpp win/pxWindowNative.cpp
             win/pxEventLoopNative.cpp win/pxTimerNative.cpp win/pxClipboardNative.cpp pxViewWindow.cpp
             win/rtMutexNative.cpp win/rtThreadPoolNative.cpp)
+	if (SUPPORT_DUKTAPE)
+      set(PXCORE_FILES ${PXCORE_FILES} rtScriptDuk/debugger/duk_debugger_interface_windows.c)
```
mfiess: Let's create two subdirectories under rtScriptDuk/debugger/.  One should be called win the other should be called unix (following the same pattern as platform specific files in pxCore/src).  Let's have both files named the same (duk_debugger_interface.c) and put the unix version under rtScriptDuk/debugger/unix/ and the windows version under rtScriptDuk/debugger/win/.

src/rtScriptDuk/debugger/duk_debugger_interface_unix.c
```diff
@@ -0,0 +1,266 @@
+#include <stdio.h>
```
mfiess: Let's rename to duk_debugger_interface.c and add it to rtScriptDuk/debugger/unix/.

src/rtScriptDuk/debugger/duk_debugger_interface_windows.c
```diff
@@ -0,0 +1,376 @@
+#undef UNICODE
```
mfiess: Let's rename to duk_debugger_interface.c and add it to rtScriptDuk/debugger/win/.

src/CMakeLists.txt
```diff
@@ -78,6 +81,9 @@ elseif (CMAKE_HOST_SYSTEM_NAME STREQUAL "Linux")
                 gles/pxClipboardNative.cpp gles/pxEGLProviderRPi.cpp gles/LinuxInputEventDispatcher.cpp pxViewWindow.cpp
                 unix/rtMutexNative.cpp unix/rtThreadPoolNative.cpp
                 rtFileCache.cpp rtHttpCache.cpp)
+         if (SUPPORT_DUKTAPE)
+          set(PXCORE_FILES ${PXCORE_FILES} rtScriptDuk/debugger/duk_debugger_interface_unix.c)
```
mfiess: Please update to the new name and location, rtScriptDuk/debugger/unix/duk_debugger_interface.c.

PR #863, opened 2018-02-13
add AppVeyor artifacts management

dwrobel:
Adds Appveyor-custodian script which manages number of builds
to make sure we will not exceed available quota.

To use it with AppVeyro CI infrastructure please add in:
  Projects->Settings->Environment->Environment variables

  API_TOKEN variable.

Token can be found on API token page under your AppVeyor account.

### DIFF
--- a/ci/build_win32.bat
+++ b/ci/build_win32.bat
@@ -25,6 +25,7 @@ call "C:/Program Files (x86)/Microsoft Visual Studio/2017/Community/VC/Auxiliary
 @rem build dependencies
 cd examples/pxScene2d/external
 call buildWindows.bat
+if %errorlevel% neq 0 exit /b %errorlevel%
 
 @rem Avoid using link.exe from that paths
 set PATH=%PATH:C:\Program Files\Git\usr\bin;=%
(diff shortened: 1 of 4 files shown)


ci/build_win32.bat
```diff
@@ -39,10 +40,26 @@ cd build-win32
...
 7z a -y pxscene-setup.zip pxscene-setup
 
 cd %ORIG_DIR%
+
+@rem deploy artifacts
+@rem based on: https://www.appveyor.com/docs/build-worker-api/#push-artifact
+if "%APPVEYOR%"=="True" (
+    if "%DEPLOY_ARTIFACTS%"=="True" (
+        @rem NSIS based installer
+        appveyor PushArtifact "build-win32\\_CPack_Packages\\win32\\NSIS\\pxscene-setup.exe" -DeploymentName "installer" -Type "Auto" -Verbosity "Normal"
```
arun-govindan: Looks like we are already inside NSIS folder in line 49. So providing the path as "build-win32\\_CPack_Packages\\win32\\NSIS\\pxscene-setup.exe" might fail. The artifact name alone should work fine.

Kindly clarify.

dwrobel: >Looks like we are already inside NSIS folder in line 49. So providing the path as "build-win32\_CPack_Packages\win32\NSIS\pxscene-setup.exe" might fail. 

It was implemented according to documentation: https://www.appveyor.com/docs/build-worker-api/#push-artifact
> `<path>`: the path to the file to upload, relative to your project’s checkout

PR #875, opened 2018-02-14
Duktape support with single build

arun-govindan:


.travis.yml
```diff
@@ -24,8 +24,9 @@ script:
 - ./ci/script.sh
 - if [ "$TRAVIS_EVENT_TYPE" = "push" ] || [ "$TRAVIS_EVENT_TYPE" = "pull_request" ]; then codecov --build "$TRAVIS_OS_NAME-$TRAVIS_COMMIT-$TRAVIS_BUILD_NUMBER" -X gcov -f $TRAVIS_BUILD_DIR/tracefile ; fi
 - if [ "$TRAVIS_EVENT_TYPE" = "push" ] || [ "$TRAVIS_EVENT_TYPE" = "pull_request" ]; then genhtml -o $TRAVIS_BUILD_DIR/logs/codecoverage $TRAVIS_BUILD_DIR/tracefile; fi
-after_script:
 - ./ci/after_script.sh
```
conniefry: Why have you moved the after_script.sh script to be executed before the travis after_script step?

arun-govindan: Ideally “after-script” is the section, where the final steps will be executed,  so initially we had the uploading of logs and removing the permission file in that section. 
Now after uploading the logs, we start running spark for duktape, which means, uploading logs are not the final step. So I moved the after script to further down. So now after script only contains removal of the permission file.

Issue #790, opened 2018-01-17
Provide linux package(s)


andrewbenton:
Current packaging provides nothing for linux.  There should be rpm/deb or flatpak/snap available for use as well as the OSX and Windows distributions.

PR #877, opened 2018-02-15
Exposes git version info to the pxscene runtime

andrewbenton:
The exposed information is derived from `git describe --tag --always
--long --dirty`, which will expose information in the form of:
`tag-offset-hash(-dirty)?`.  This is broken then loaded into the
scene.info.build variable as:
  - major
  - minor
  - micro
  - nano
  - tag
  - offset
  - hash
  - dirty

The major / minor / micro / nano are formed from the four period
separated values in the tag, or zeros if it can't be parsed.

In the case that the GIT_PXSCENE_VERSION variable isn't provided, it
will default to `0.0.0.0-0-0000000000-dirty`.

This will help developers to determine the environment that their
application has run against down to the commit.

### DIFF
--- a/examples/pxScene2d/src/CMakeLists.txt
+++ b/examples/pxScene2d/src/CMakeLists.txt
@@ -287,6 +287,21 @@ else ()
     add_definitions(-DPX_SCENE_VERSION=${PXSCENE_VERSION})
 endif("${PXSCENE_VERSION}" STREQUAL "")
 
+execute_process(
+  COMMAND git describe --tags --always --long --dirty
+  WORKING_DIRECTORY "${CMAKE_CURRENT_SOURCE_DIR}"
+  RESULT_VARIABLE GIT_DESCRIBE_RES
+  OUTPUT_VARIABLE GIT_PXSCENE_VERSION
+  ERROR_QUIET
+  OUTPUT_STRIP_TRAILING_WHITESPACE
+)
+
+if(GIT_DESCRIBE_RES EQUAL 0)
+  message("Settings pxscene version to: ${GIT_PXSCENE_VERSION}")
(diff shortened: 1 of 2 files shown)


examples/pxScene2d/src/CMakeLists.txt
```diff
@@ -287,6 +287,21 @@ else ()
     add_definitions(-DPX_SCENE_VERSION=${PXSCENE_VERSION})
 endif("${PXSCENE_VERSION}" STREQUAL "")
 
+execute_process(
+  COMMAND git describe --tags --always --long --dirty
```
dwrobel: This is a great idea but surprisingly against your https://github.com/pxscene/pxCore/issues/790 request.
Please note that for #790 a bare tarball is necessary and this will be no longer Git repository _(.git sub directory)_. So using the propsed solution we will end up in 0.0.0.0-0-0000000000-dirty version all the time.

Reading a version from a file (which can be included in a tarball) might be a solution which will work in both cases.

andrewbenton: I like the idea of generating a version.txt at configuration / compile time, possibly from the `git describe` if it exists.

About the tarball comment, what is that required for?  It doesn't immediately seem necessary from when I was looking at CPack.

dwrobel: >About the tarball comment, what is that required for?

If any Linux distribution will be interested in providing native package e.g. .rpm or .deb the only thing they really need is a tarball from https://github.com/pxscene/pxCore/releases. Unfortunately executing git command within this unpacked tarball will simply not return the proper software version.

mfiess:
A new solution is in the works

PR #886, opened 2018-02-18
Correct breakpad patching, simpler ucontext patch

andrewbenton:
The breakpad patching wasn't working, at the very least, because the
patch path assumed that quilt was being run from outside of the breakpad
directory.  The patch path has been corrected.

The patch that is now applied is also simpler now, as it only entails
the addition of six lines that should work whether ucontext or
ucontext_t are being provided by the glibc API.

Supersedes #742 and tested on Fedora Linux 27, which provides 
ucontext_t only.


dwrobel:
>patch path assumed that quilt was being run from outside of the breakpad

The patch is actually applied from `breakpad` directory (see below):
https://github.com/dwrobel/pxCore/blob/master/examples/pxScene2d/external/build.sh#L147

I'm using it on my F27 and it works fine. Could you please provide any compilation error log, for that issue?

mfiess:
Closing based on Damian's comment.  Let's reopen if this is still an issue.

PR #890, opened 2018-02-20
Fix compilation issue : Usage of __PRETTY_FUNCTION__ causes compilati…

arun-govindan:
Fix for compilation issue : Usage of __PRETTY_FUNCTION__ causes compilation issue in windows.


arun-govindan:
PR already raised

PR #904, opened 2018-02-21
Fix for crash in pxWayland::launchAndMonitorClient

GlebNovodranPE:
Fix for crash upon going to standby or shutdown, caused by pxWayland::launchAndMonitorClient running in another thread and using the freed context pointer. The per-cautionary checks within the Westeros routines seems to don't because the pointer is not zero.

### DIFF
--- a/examples/pxScene2d/src/pxWayland.cpp
+++ b/examples/pxScene2d/src/pxWayland.cpp
@@ -91,6 +91,7 @@ pxWayland::~pxWayland()
   if ( mWCtx )
   {
      WstCompositorDestroy(mWCtx);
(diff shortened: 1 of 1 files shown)


examples/pxScene2d/src/pxWayland.cpp
```diff
@@ -91,6 +91,7 @@ pxWayland::~pxWayland()
   if ( mWCtx )
   {
      WstCompositorDestroy(mWCtx);
+     mWCtx = 0;
```
mfiess: Can you please change to mWCtx = NULL?

PR #923, opened 2018-02-27
Appveyor cron build

arun-govindan:


ci/build_win32.bat
```diff
@@ -53,7 +53,11 @@ cd %ORIG_DIR%
 
 @rem deploy artifacts
 @rem based on: https://www.appveyor.com/docs/build-worker-api/#push-artifact
-if "%APPVEYOR%"=="True" (
+set cronUpload=Fasle
+
+if "%APPVEYOR_SCHEDULED_BUILD%"=="True" if  "%APPVEYOR_REPO_NAME%" = "pxscene/pxcore" if "%APPVEYOR_REPO_BRANCH%" == "master"  set cronUpload=True
+
+if %cronUpload%==True (
```
dwrobel: This will prevent people using forks to keep/publish Artifacts.

arun-govindan: Yes , I agree. I thought the binary will be released as edge builds. So only master's code should be used.
But yeah the artifacts storage is account specific. I will keep only APPVEYOR_SCHEDULED_BUILD condition (because our current need is to only upload artifacts from scheduled builds) and remove the other two.

PR #930, opened 2018-03-01
Fixes for NULL pointer dereference

FitzerIRL:


src/mac/pxClipboardNative.mm
```diff
@@ -12,22 +12,35 @@
 
 std::string pxClipboardNative::getString(std::string type)
 {
-    printf("pxClipboardNative::getString() - ENTER\n");
+//    printf("pxClipboardNative::getString() - ENTER\n");
     
     NSPasteboard*  pasteBoard = [NSPasteboard generalPasteboard];
     NSString*        myString = [pasteBoard  stringForType: NSPasteboardTypeString];
```
FitzerIRL: NOTE:  An if( ) checking 'pasteBoard' is not explicitly necessary.  ObjectiveC ignores attempts to send messages to NULL objects ....    Where  the ObjectiveC syntax   [obj message]  ... sends the 'obj' a message to call 'message'

PR #932, opened 2018-03-05
fix for crash on simultaneous child dispose

madanagopalt:


conniefry:
Please be sure to add tests for this change.

PR #944, opened 2018-03-08
Windows compatibility fix for the dukluv modules lookup hack

sergiygladkyy:
Replaced std::string usage in a C code causing Windows build to fail


sergiygladkyy:
Just realized that there is a patch similar to mine: `examples/pxScene2d/external/dukluv/patches/dukluv.git.patch`

But `main.c`  hasn't been patched on Windows. So, I think, the pull request can be disregarded, I will look into the patching problem

PS. Why do we have that weird c++ code in the original main.c in the first place?

PR #949, opened 2018-03-12
Updated ELLIPSIS_STR with the hexa equivalent of its unicode value

arun-govindan:


arun-govindan:
Raised a different PR

PR #958, opened 2018-03-14
Update license

arun-govindan:
Directories included for scanning : 
~/pxCore_arunG/pxCore//src
~/pxCore_arunG/pxCore//remote
~/pxCore_arunG/pxCore//tests
~/pxCore_arunG/pxCore//examples/pxScene2d/src/
Files extensions included for scanning : C, CPP, H, JS

Note : Files which does not contain Apache license are searched for string "copyright" and "Licensed under the", if either one presents, then the file is excluded assuming it has another valid license, for example MIT, GPL license


conniefry:
Source in examples/pxScene2d/src/node_modules should have their licenses untouched.  These are modules not owned by us.

PR #964, opened 2018-03-15
Added loseFocusChain bool property for onBlur Event (XRE2-1268)

anand-ky:


conniefry:
Ideally, there will be a test added to tests.json as part of this PR to ensure validation of this new feature in CI builds.

PR #968, opened 2018-03-16
fix mData NULL deref in rtString::append

ptrekels:


conniefry:
Thanks, Peter.  Could you please add a small unit test to demonstrate this problem and to test the fix?

PR #972, opened 2018-03-20
testing builds

madanagopaltcomcast:


examples/pxScene2d/src/pxScene.cpp
```diff
@@ -270,7 +276,15 @@ class sceneWindow : public pxWindow, public pxIViewContainer
...
 
     context.term();
+    script.pump();
     script.collectGarbage();
+    #ifndef WIN32
+    sleep(5);
+    #endif
+    gUIThreadQueue.process(1.0);
+    #ifndef WIN32
+    sleep(5);
```
mfiess: Let's try to remove these sleeps since all the javascript is processed on this thread.

examples/pxScene2d/src/pxScene.cpp
```diff
@@ -270,7 +276,15 @@ class sceneWindow : public pxWindow, public pxIViewContainer
...
+    #endif
 
     if (gDumpMemUsage)
     {
```
mfiess: Let's move the extra code above to inside the `if (gDumpMemUsage)` block.

PR #983, opened 2018-03-21
time related functions - implementation cleanup

dwrobel:
time related functions - implementation cleanup
    
     - introduces the new common implementation of time-related functions
       for all platforms supporting c++11,
     - try to synchronize legacy (non c++11) time implementations,
     - fixes legacy (non c++11) implementation of pxSleepMS(),
     - introduces new: void pxSleepUS(uint64_t usToSleep); function,
    
     - fixes pxTimerNativeTest failure:
        /Users/travis/build/pxscene/pxCore/tests/pxScene2d/test_pxTimerNative.cpp:37: Failure
        Value of: (int)(endTime - startTime) == TEST_TIME
          Actual: false
        Expected: true
        [  FAILED  ] pxTimerNativeTest.pxTimerNativeTestS (2000 ms),
    
     - start using pxSleep*() in wayland_egl/pxWindowNative.cpp backend,

     - even pacing on wayland_egl backend.


mfiess:
Let's leave the original file unmodified, but we can have a cmake option to using the new C++11 syntax (chrono) if configured.  Otherwise we use the current files (as is).  We can discuss in our next meeting.

PR #989, opened 2018-03-22
coverity fixes

madanagopaltcomcast:


remote/rapidjson/internal/diyfp.h
```diff
@@ -41,7 +41,7 @@ RAPIDJSON_DIAG_OFF(padded)
 #endif
 
 struct DiyFp {
-    DiyFp() {}
+    DiyFp() : f(0), e(0) {}
```
mfiess: We should avoid direct edits for externals and rapidjson.  Please update with with a patch file.

PR #990, opened 2018-03-22
fix 'strict-aliasing' compilation errors and typo in pxClamp usage

dwrobel:
Compiler output:
  pxCore/examples/pxScene2d/src/rasterizer/xs_code/xs_Float.h: In function ‘int32 xs_CRoundToInt(real64, real64)’:
  pxCore/examples/pxScene2d/src/rasterizer/xs_code/xs_Float.h:349:33: error: dereferencing type-punned pointer will break strict-aliasing rules [-Werror=strict-aliasing]
    return ((int32*)&val)[_xs_iman_];
                                 ^
Note: The 'xs_code' seems to be complete broken e.g.

  typedef long uint32; // <- uint32 will not have 32-bits on x86-64!!!

then uses something like the following:

  double val;
  return ((int32*)&val)[1]; // <- on 64-bit Big Endian platforms it will read random memory

thus it seems to be simpler to not use 'xs_code' at all.

The fix adds XS_CODE_ENABLED which, by default do not use implementation
from 'xs_code' and, provides equivalent implementation of two function
which were effectively used from the 'xs_code':

  int32_t xs_RoundToInt(const double d);
  int32_t xs_CRoundToInt(const double d);

### DIFF
--- a/examples/pxScene2d/src/rasterizer/pxRasterizer.cpp
+++ b/examples/pxScene2d/src/rasterizer/pxRasterizer.cpp
@@ -1,9 +1,25 @@
 #include "pxTimer.h"
 #include "pxRasterizer.h"
 
-#include "xs_Core.h"
-#include "xs_Float.h"
+#ifdef XS_CODE_ENABLED
+#  include "xs_Core.h"
+#  include "xs_Float.h"
+#else
+#  include <cstdint>
+#  include <math.h>
+   inline int32_t xs_RoundToInt(const double d)
+   {
+     return floor(d + 0.5);
+   }
+
+   inline int32_t xs_CRoundToInt(const double d)
+   {
+     return ceil(d);
+   }
+#endif
+
 //#include "rtLog.h"
+
 
 #include <algorithm>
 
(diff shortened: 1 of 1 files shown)


examples/pxScene2d/src/rasterizer/pxRasterizer.cpp
```diff
@@ -1,10 +1,26 @@
...
+   inline int32_t xs_RoundToInt(const double d)
+   {
+     return floor(d + 0.5);
+   }
+
+   inline int32_t xs_CRoundToInt(const double d)
+   {
+     return ceil(d);
+   }
+#endif
```
dwrobel: One comment related to this implementation:

- xs_RoundToInt() is implemented as per https://github.com/pxscene/pxCore/blob/master/examples/pxScene2d/src/rasterizer/xs_code/xs_Float.h#L403

- xs_CRoundToInt() original implementation is at https://github.com/pxscene/pxCore/blob/master/examples/pxScene2d/src/rasterizer/xs_code/xs_Float.h#L352

For `_xs_DEFAULT_CONVERSION != 0` both implementations looks the same, so why there were two different functions? I simply guessing that maybe there is an issue and assumed that `C` in `xs_CRoundToInt` stands for `ceil()` and implemented it in this way. If that assumption is entirely wrong, please let me know.

FitzerIRL: Looks ok to move towards removal.

PR #1028, opened 2018-04-03
Appveyor builds

arun-govindan:


ci/appveyor-custodian.py
```diff
@@ -73,7 +73,7 @@ def get_env_or_exit(name, default_value = None, exit_code = 0):
     builds_num = len(builds["builds"])
     logging.info("Builds: %d" % (builds_num))
 
-    keep_builds_num = int(get_env_or_exit('BUILDS_TO_KEEP', default_value = 20))
+    keep_builds_num = int(get_env_or_exit('BUILDS_TO_KEEP', default_value = 50))
```
dwrobel: It's fully configurable using env. variable so why do we need to change default especially if  50 is likely not to work, please compute 50 x 2 artifacts x 10MB each artifact gives you 1000MB which is equal to the quota available on the AppVeyour service.

arun-govindan: Hi Damian,
We are not storing artifacts for all the build.  Only for edge builds we are storing, which runs twice a day.
In that case, if there are 50 builds, builds which are triggered by PR and Push will not have artifacts.
And also the builds with artifacts will be get deleted, eventually. 
So i believe the total size will not be reached. And so instead of setting the env, I updated the default value.

dwrobel: > In that case, if there are 50 builds, builds which are triggered by PR and Push will not have artifacts.

That doesn't matter.

> Only for edge builds we are storing, which runs twice a day.

That means that after 25 days you will reach 50 builds stored on the AppVeyor anyway and based on my experience it won't work. If you are curious to just check it, please consider to change env. variable first.

arun-govindan: Hi,
In the second commit 482bc01, I have already reduced the count to 40 a week ago, but still i am thinking it to reduce to 35. I will configure it through the env variable and revert the above modification.
Thanks.

dwrobel: One thing which comes to my mind, bear in mind that once you exceed the AppVeyor quota, all the builds *(even those ones which don't store any artifacts will start failing)*, see: https://help.appveyor.com/discussions/problems/11202-the-remote-server-returned-an-error-500-internal-server-error-on-push-artifact

arun-govindan:
Closing this PR as this requirement is addressed in a different way.

PR #1031, opened 2018-04-03
Added loseFocusChain bool property for onBlur Event

anand-ky:
Added JS test case in tests.json to verify that loseFocusChain is set.


anand-ky:
Going to submit new PR addressing review comments from Mike.

PR #1044, opened 2018-04-06
Fix for ellipisis not rendered in windows

arun-govindan:


arun-govindan:
Raised a different PR

PR #1059, opened 2018-04-11
flickering fix

dwrobel:


examples/pxScene2d/src/pxScene2d.cpp
```diff
@@ -2318,7 +2318,7 @@ ENTERSCENELOCK()
     mRoot->drawInternal(true);
 EXITSCENELOCK()
     context.popState();
-    mLastFrameDirtyRect.setLTRB(mDirtyRect.left(), mDirtyRect.top(), mDirtyRect.right(), mDirtyRect.bottom());
+    mLastFrameDirtyRect.setLTRB(dirtyRectangle.left(), dirtyRectangle.top(), dirtyRectangle.right(), dirtyRectangle.bottom());
```
mfiess: This change will disable dirty rectangles.  The dirty rect will always be full screen since the previous dirty rectangle (mLastFrameDirtyRect) will always be carried over.

Issue #1064, opened 2018-04-11
ThreadSanitizer(fishtank.js): heap-use-after-free @pxScene2d.cpp:2923 in pxScene2d::bubbleEventOnBlur(rtObjectRef, rtRef<pxObject>, rtRef<pxObject>)


dwrobel:
pxScene compiled with -fsanitize=thread and run with https://raw.githubusercontent.com/pxscene/pxscene/gh-pages/examples/px-reference/gallery/fishtank/fishtank.js gives the following output:

```
WARNING: ThreadSanitizer: heap-use-after-free (pid=8791)
  Read of size 8 at 0x7b04000047c8 by main thread (mutexes: write M368):
    #0 operator==(rtRef<pxObject> const&, rtRef<pxObject> const&) <null> (pxscene+0x00000057e5f2)
    #1 pxScene2d::bubbleEventOnBlur(rtObjectRef, rtRef<pxObject>, rtRef<pxObject>) /home/sw/projects/pxscene/pxCore/examples/pxScene2d/src/pxScene2d.cpp:2923 (pxscene+0x00000055954a)
    #2 pxScene2d::setFocus(rtObjectRef) /home/sw/projects/pxscene/pxCore/examples/pxScene2d/src/pxScene2d.cpp:2779 (pxscene+0x0000005581e1)
    #3 pxObject::setFocus(bool) /home/sw/projects/pxscene/pxCore/examples/pxScene2d/src/pxScene2d.cpp:658 (pxscene+0x00000054c51d)
    #4 pxObject::setFocus_PropSetterThunk(rtValue const&) /home/sw/projects/pxscene/pxCore/examples/pxScene2d/src/pxScene2d.h:186 (pxscene+0x000000564027)
    #5 rtObject::Set(char const*, rtValue const*) /home/sw/projects/pxscene/pxCore/src/rtObject.cpp:401 (pxscene+0x000000628783)
    #6 pxObject::Set(char const*, rtValue const*) /home/sw/projects/pxscene/pxCore/examples/pxScene2d/src/pxScene2d.cpp:686 (pxscene+0x00000054c72b)
    #7 void rtScriptNodeUtils::rtObjectWrapper::setProperty<char const*>(char const* const&, v8::Local<v8::Value>, v8::PropertyCallbackInfo<v8::Value> const&) /home/sw/projects/pxscene/pxCore/src/rtScriptNode/rtObjectWrapper.cpp:229 (pxscene+0x0000005f5f10)
[...]

PR #1067, opened 2018-04-11
fix incorrect memory access

mfiess:


dwrobel:
Great news! It fixes for me #1064.

PR #1088, opened 2018-04-18
updates for wayland app termination

mfiess:


conniefry:
Bug fix for slow wayland app shutdowns.

PR #1089, opened 2018-04-18
Added support to Optimus to accept uri for wpe and spark applications

binuinbaraj:


examples/pxScene2d/src/rcvrcore/optimus.js
```diff
@@ -34,8 +35,15 @@ function Application(props) {
   }
   if (cmd){
     if (scene !== undefined) {
-      if (cmd.includes(".js")){
-        this.externalApp = scene.create({t:"scene", parent:root, url:cmd});
+      if (cmd.includes(".js") || (cmd == "spark")){
+        if (cmd.includes(".js")){
```
mfiess: I think we can remove the `cmd.includes(".js")` check to force a cut over to the new method of launching Spark apps.

PR #1100, opened 2018-04-21
test builds

madanagopalt:


src/rtScriptNode/rtScriptNode.cpp
```diff
@@ -1035,27 +1035,36 @@ rtError rtScriptNode::pump()
...
 //#else
-  Locker                locker(mIsolate);
-  Isolate::Scope isolate_scope(mIsolate);
-  HandleScope     handle_scope(mIsolate);    // Create a stack-allocated handle scope.
+  // found a problem where if promise triggered by one event loop gets resolved by other event loop.
+  // It is causing the dependencies between data running between two event loops failed, if one one 
+  // loop didn't complete before other. So, promise not registered by first event loop, before the second
+  // event looop sends back the ready event
+  static bool isPumping = false;
+  if (isPumping == false)
```
mfiess: Please perform this check only if RUNINMAIN is defined.

PR #1105, opened 2018-04-25
validate javascript execution on compile optimizations

madanagopalt:


madanagopalt:
change looks generic and not suggested to be safe from v8 team. So, ignoring this

PR #1109, opened 2018-04-25
Changes to add rdkbrowser2 support to optimus

binuinbaraj:


mfiess:
The proper support was added with PR 1152.

PR #1114, opened 2018-04-26
fix for permissions interfering with service bubbling within a scene's hierarchy

mfiess:


conniefry:
Change looks good to me, but the unit test needs to be changed accordingly, as well.

PR #1123, opened 2018-05-03
use no origin for file://

npoltorapavlo:


src/rtUrlUtils.cpp
```diff
@@ -91,21 +90,30 @@ rtString rtUrlEncodeParameters(const char* url)
...
     // URI = scheme ":" hier-part [ "?" query ] [ "#" fragment ]
     // scheme = ALPHA *( ALPHA / DIGIT / "+" / "-" / "." )
-    const char* u = url;
+    u = url;
     for (; *u && (*u == '+' || *u == '-' || *u == '.' || isalpha(*u) || isdigit(*u)); u++);
     if (*u == ':' && u != url)
     {
       u++;
-      if (0 == strncmp(u, "//", 2))
+      if (*u == '/' && *(u + 1) == '/')
```
mfiess: Why the change from strncmp?

npoltorapavlo: 2 symbols can be checked without string.h . I also replaced strcmp with rtString's "compare" in rtCORSUtils.cpp

PR #1128, opened 2018-05-04
CORS/permissions runtime checks, renamings

npoltorapavlo:


ci/execute_osx.sh
```diff
@@ -33,7 +33,9 @@ dumped_core=0
 
 export PX_DUMP_MEMUSAGE=1
 export RT_LOG_LEVEL=info
-export PXSCENE_PERMISSIONS_CONFIG=$TRAVIS_BUILD_DIR/examples/pxScene2d/src/pxscenepermissions.conf
+export SPARK_CORS_ENABLED=false
```
mfiess: We should enable CORS by default for all builds.

ci/execute_linux.sh
```diff
@@ -32,6 +32,9 @@ export VALGRINDLOGS=$TRAVIS_BUILD_DIR/logs/valgrind_logs
 export PX_DUMP_MEMUSAGE=1
 export ENABLE_VALGRIND=1
 export RT_LOG_LEVEL=info
+export SPARK_CORS_ENABLED=false
```
mfiess: We should enable CORS by default for all builds.

src/rtCORSUtils.cpp
```diff
@@ -20,38 +20,50 @@
...
+    didCheck = true;
+    const char* s = getenv(ENV_NAME_ENABLED);
+    if (s != NULL)
+    {
+      rtString envVal(s);
+      enable = 0 == envVal.compare("true") || 0 == envVal.compare("1");
+    }
+    else
+    {
+      enable = false;
```
mfiess: Let's enable CORS by default.  The environment variable can be used to disable it.

PR #1129, opened 2018-05-04
coverity fix for texture memory limit

mfiess:


mfiess:
The fix for this is included in PR 1138.

PR #1133, opened 2018-05-07
Fix for ellipsis rendering issue in windows

arun-govindan:


arun-govindan:
Tticket already resolved

PR #1136, opened 2018-05-07
Enabling thread sanitizer on travis

arun-govindan:


.travis.yml
```diff
@@ -4,7 +4,7 @@ matrix:
     dist: trusty
     sudo: required
   - os: osx
-    osx_image: xcode6.4
+    osx_image: xcode9.3
```
mfiess: Will switching to xcode 9.3 break our compatibility with older versions of macOS?

PR #1141, opened 2018-05-08
nuke unused xs_code (follow-up: #990)

dwrobel:


dwrobel:
>What is the Need/Motivation ?
>Saving space ?

Mainly to prevent others start using any stuff from `xs_code`.

PR #1167, opened 2018-05-14
add & use components' specific build.sh scripts

dwrobel:


dwrobel:
We will make use of those scripts just to build `dukluv` and `breakpad` components only.

PR #1172, opened 2018-05-15
Avoid clearing promise queue when event loop is running

madanagopaltcomcast:


src/rtScriptNode/rtScriptNode.cpp
```diff
@@ -95,7 +95,9 @@ using namespace rtScriptNodeUtils;
 #ifndef DISABLE_USE_CONTEXTIFY_CLONES
 # define USE_CONTEXTIFY_CLONES
 #endif
-
+#ifdef RUNINMAIN
+bool isPumping = false;
```
mfiess: Let's rename to `isPumping` to `gIsPumpingJavaScript`

PR #1173, opened 2018-05-15
remove pump call from dispose

kannanmadanagopalt:


kannanmadanagopalt:
closing this at this time, as this seems to have side-effects on apps using js engine resources in their onClose event.

PR #1180, opened 2018-05-16
add exit test to testRunner

dwrobel:


dwrobel:
This is just a proposal. The preferred solution would be to move the `exitTest.js` to the place other tests are located.

conniefry:
@dwrobel The test is now published as https://px-apps.sys.comcast.net/pxscene-samples/examples/px-reference/tests/test_exit.js, so you should be able to use a relative path in the tests.json file.

dwrobel:
> so you should be able to use a relative path in the tests.json file.

Thank you, PR has been updated accordingly.

dwrobel:
Conflict resolved.

dwrobel:
Switch to use v7 which should solve a build failure mentioned https://github.com/pxscene/pxCore/pull/1180#issuecomment-399062786.

PR #1209, opened 2018-05-30
have optimus call destroy on spark app

mfiess:


mfiess:
The updated version is in PR 1232.

PR #1213, opened 2018-06-01
 Fix a crash during rtRemote Server shutdown

apaschenko:
- Removing all invalid objects from ObjectCache during the shutdown of any client to prevent crash when the server is shut down;
- Fix a typo.

### DIFF
--- a/remote/rtRemoteServer.cpp
+++ b/remote/rtRemoteServer.cpp
@@ -443,6 +443,8 @@ rtRemoteServer::onClientStateChanged(std::shared_ptr<rtRemoteClient> const& clie
     }
   }
 
+  m_env->ObjectCache->removeUnused(true);
+
   return e;
 }
 
--- a/remote/rtRemoteObjectCache.cpp
+++ b/remote/rtRemoteObjectCache.cpp
@@ -197,7 +197,7 @@ rtRemoteObjectCache::erase(std::string const& id)
 }
 
 rtError
-rtRemoteObjectCache::removeUnused()
+rtRemoteObjectCache::removeUnused(bool expireAll)
 {
   auto now = std::chrono::steady_clock::now();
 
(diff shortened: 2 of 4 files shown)


examples/pxScene2d/src/pxWayland.cpp
```diff
@@ -88,6 +88,7 @@ pxWayland::~pxWayland()
 {
 #ifdef ENABLE_PX_WAYLAND_RPC
   rtRemoteUnregisterDisconnectedCallback(pxWayland::remoteDisconnectedCB, this);
+  rtRemoteShutdown();
```
mfiess: Jake, can you please comment on if it's okay to call rtRemoteShutdown() after destroying each pxWayland object?

Gavriliuk:
The goal of this change is to solve the problem which Jake and Mike talk about.

apaschenko:
@mfiess Thank you, Michael. Ok, I'll try to find another way for the fixing of group rtRemoteShutdown/rtRemoteInit bugs.

apaschenko:
@mfiess I've removed rtRemoteShutdown() from pxWayland's destructor.

remote/rtRemoteServer.cpp
```diff
@@ -443,6 +443,8 @@ rtRemoteServer::onClientStateChanged(std::shared_ptr<rtRemoteClient> const& clie
     }
   }
```
gladish: Client seems to have states are ok also 
  enum class State
  {
    Started,
    Shutdown
  };

Shouldn't this be guarded with if (state == Shutdown) ?

apaschenko: Hi @gladish ,
This code is not on the client but on the server side (rtRemoteServer) and it is called only when the client initiates the "Shutdown" event.

mfiess:
The rtRemote code in pxCore will no longer be used very shortly.  Please submit this change to the rtRemote repository: https://github.com/pxscene/rtRemote.  All rtRemote changes will go to this repository going forward.

PR #1235, opened 2018-06-11
sending ready promise for textbox and text elements without parent

madanagopalt:
please review


examples/pxScene2d/src/pxTextBox.cpp
```diff
@@ -111,8 +117,7 @@ void pxTextBox::onInit()
   // If this is using the default font, we would not get a callback
   if(mFontLoaded || (getFontResource() != NULL && getFontResource()->isFontLoaded()))
   {
-    mFontLoaded = true;
-    setNeedsRecalc(true);
+    resourceReady("resolve");
```
conniefry: The textBox promise cannot be resolved here.  Measurements must be calculated and only then can the promise be resolved.

madanagopalt:
Please review the new changes

conniefry:
What about a new testRunner/CI test to ensure this is working correctly?

PR #1238, opened 2018-06-12
Remove pxPath from the build

FitzerIRL:
Adding '#ifdef ENABLE_PXPATH'  to remove pxPath from the build


examples/pxScene2d/src/rasterizer/pxPath.cpp
```diff
@@ -18,6 +18,9 @@
 
 #include "pxScene2d.h"
 #include "pxContext.h"
+
+#ifdef ENABLE_PXPATH
```
mfiess: Let's remove the need to surround this whole file with `#ifdef ENABLE_PXPATH`.  Let's simply not include pxPath.cpp as part of the build if the cmake option is not defined.

FitzerIRL: Migrated to CMake approach to remove pxPath via 'BUILD_WITH_PXPATH=OFF' (default) build switch

Code pushed.

PR #1249, opened 2018-06-18
Add access permission checks for local and remote file accesses

madanagopalt:


examples/pxScene2d/src/pxScene2d.cpp
```diff
@@ -3607,6 +3608,18 @@ rtError pxScene2d::checkAccessControlHeaders(const rtString& url, const rtString
 #endif
 }
 
+rtError pxScene2d::sparkSetting(const rtString& setting, rtValue& value) const
+{
+  rtValue val;
+  if (RT_OK != rtSettings::instance()->value(setting, val))
+  {
+    value = rtString("SETTINGS_NOT_FOUND");
```
mfiess: Let's return a null value if the setting is not found.

ci/supportfiles/.sparkSettings.json
```diff
@@ -0,0 +1 @@
+{"useSparkPermissionsForPxFileAccess" : true}
```
mfiess: There shouldn't be any spark settings file by default.  This setting file should only be used to override default setting values.  Please remove this file.

ci/execute_linux.sh
```diff
@@ -195,4 +195,77 @@ else
 	checkError $retVal "Valgrind execution reported memory leaks" "$errCause" "Follow the steps locally : export ENABLE_VALGRIND=1;export SUPPRESSIONS=<pxcore dir>/ci/leak.supp;./pxscene.sh $TESTRUNNERURL?tests=<pxcore dir>/tests/pxScene2d/testRunner/tests.json and fix the leaks"
 	exit 1;
 fi
+
+
+#add test for getFile failure for localhost and pxscene.org
+cd $TRAVIS_BUILD_DIR/examples/pxScene2d/src
+mv sparkpermissions.conf sparkpermissions.conf_orig
+cp $TRAVIS_BUILD_DIR/ci/supportfiles/sparkpermissions.conf .
+cp $TRAVIS_BUILD_DIR/ci/supportfiles/.sparkSettings.json ~/.
```
mfiess: Please remove the sparkSettings file from the install.

examples/pxScene2d/src/rcvrcore/AppSceneContext.js
```diff
@@ -80,6 +81,9 @@ function AppSceneContext(params) {
   this.timers = [];
   this.timerIntervals = [];
   this.webSocketManager = null;
+  this.httpwrap = new http_wrap(this.accessControl);
+  this.httpswrap = new https_wrap(this.accessControl);
+  this.useSparkPermissionForFileAccess = this.innerscene.sparkSetting("useSparkPermissionsForPxFileAccess")
```
mfiess: Please add a check to see if this setting is defined.  Let's change the setting name to "disableFilePermissionCheck".  The default value (if the setting is not found) should be false.  This means that we should honor permissions by default.

ci/supportfiles/sparkpermissions.conf
```diff
@@ -0,0 +1,56 @@
...
+    "*://[::1]:*" : "local",
+    "*://[0:0:0:0:0:0:0:1]" : "local",
+    "*://[0:0:0:0:0:0:0:1]:*" : "local",
+    "file://*" : "local",
+    "*://www.pxscene.org" : "pxscene.org",
+    "*://www.pxscene.org:*" : "pxscene.org",
+    "*://pxscene.org" : "pxscene.org",
+    "*://pxscene.org:*" : "pxscene.org"
+  }
+}
```
mfiess: Let's avoid committing a default permissions file.  We can use the default permissions that are in code today.

PR #1260, opened 2018-06-20
Rt remote websocket

jmgasper:
Websocket support, including test HTML page


jmgasper:
#1320 is the one to use here.

PR #1264, opened 2018-06-21
Adding scaling parameters to [ imageResource ] for SVG

FitzerIRL:


src/pxUtil.h
```diff
@@ -99,8 +99,8 @@ pxImageType;
 
 pxImageType getImageType( const uint8_t* data, size_t len );
 
-rtError pxLoadImage( const char* imageData, size_t imageDataSize, pxOffscreen& o, int w = 0, int h = 0);
-rtError pxLoadImage( const char* filename,                        pxOffscreen& b, int w = 0, int h = 0);
+rtError pxLoadImage( const char* imageData, size_t imageDataSize, pxOffscreen& o, int32_t w = 0, int32_t h = 0, float sx = 1.0f, float sy = 1.0f);
+rtError pxLoadImage( const char* filename,                        pxOffscreen& b, int32_t w = 0, int32_t h = 0, float sx = 1.0f, float sy = 1.0f);
```
mfiess: The function declaration for `pxLoadImage` has been updated to take two floats.  However, the definition in pxUtil.cpp was not updated (`rtError pxLoadImage(const char *filename, pxOffscreen &b, int32_t w /* = 0*/, int32_t h /* = 0*/)`).  This leads to the linker error.

FitzerIRL: Thanks for catching this.   Fixed

PR #1271, opened 2018-06-22
Changes for clearing resource promises pending in downloads

madanagopalt:


examples/pxScene2d/src/pxResource.cpp
```diff
@@ -51,6 +51,12 @@ pxResource::~pxResource()
   {
     gUIThreadQueue->removeAllTasksForObject(this);
   }
+  if(((rtPromise*)mReady.getPtr())->status() == false)
```
mfiess: Is it possible for a pxResource to be destroyed if JavaScript code is still waiting on its promise?

conniefry: I would think not.

PR #1294, opened 2018-07-02
scene suspend and resume support

mfiess:


AKMK12345:
How to draw a circle using pxcore.
Or how to create a new object in scene.create such as circle?

conniefry:
This question has been moved to new Issue #1457 since it has nothing to do with this Pull Request.

PR #1297, opened 2018-07-03
Fix for crash while spark exits

arun-govindan:


arun-govindan:
Actually Mike, there is a issue in Mac travis build alone, as gcda files for pxscene is not getting generated. So i tried commenting the fixed code to check whether it runs fine or not.
And the issue happens only when the code is added. I am working on to check where the issue is (the issue is not reproducable locally).

PR #1298, opened 2018-07-04
rtRemote changes for lower versions of gcc

madanagopaltcomcast:


mfiess:
The roadmap for rtRemote has changed, and we no longer need support for older versions of C++ toolchains.  Closing this PR for now.

PR #1312, opened 2018-07-09
Make move a no op if child is already in that position

conniefry:
Existing tests already cover this scenario.


conniefry:
Reopening to trigger new build.

PR #1322, opened 2018-07-12
redesign CORS implementation

npoltorapavlo:
- move implementation to rtCORS class (replaces pxScene2d::checkAccessControlHeaders)
- implement http headers parsing
- check Access-Control-Allow-Credentials
- apply CORS to images, fonts
- build rtCORS no matter enabled or not (reduce amount of ifdef-s)
- remove unnecessary build time switch PXCORE_ACCESS_CONTROL_CHECK
- remove unnecessary build time switch PXSCENE_TEST_ACCESS_CONTROL_CHECK
- adjust the unit test test_cors.cpp for the new rtCORS
- fix warning in unit test test_rtSettings.cpp
- adjust CI list tests.json: one test used cross-origin for the font


examples/pxScene2d/src/pxScene2d.cpp
```diff
@@ -1926,6 +1923,9 @@ pxScene2d::pxScene2d(bool top, pxScriptView* scriptView)
   // rtPermissions accounts parent scene permissions too
   mPermissions = new rtPermissions(mOrigin.cString());
 #endif
+#ifdef ENABLE_ACCESS_CONTROL_CHECK
+  mCORS = new rtCORS(mOrigin.cString());
+#endif
```
mfiess: It looks like ENABLE_ACCESS_CONTROL_CHECK will no longer be defined.  Should we remove this block of code or is the intention to always setup CORS here (or perhaps some other logic)?

npoltorapavlo: Previously ENABLE_ACCESS_CONTROL_CHECK was defined in 3 different CMakeLists.txt and all CORS related code was wrapped in ifdef-s. Now it is used only in this place. If turned off mCORS will be NULL making it clear that CORS is off. By default it is on in examples/pxScene2d/src/CMakeLists.txt

PR #1338, opened 2018-07-24
Fix for OSCILLATE bug

FitzerIRL:


FitzerIRL:
Prefer this PR - https://github.com/pxscene/pxCore/pull/1299

PR #1352, opened 2018-07-30
Fix for travis build not getting stopped on a crash.

arun-govindan:


ci/unittests_linux.sh
```diff
@@ -24,16 +26,31 @@ TESTLOGS=$TRAVIS_BUILD_DIR/logs/test_logs;
 grep "Global test environment tear-down" $TESTLOGS
 retVal=$?
 count=0
-while [ "$retVal" -ne 0 ] &&  [ "$count" -ne 180 ]; do
+corefile=1
+while [ "$retVal" -ne 0 ] &&  [ "$count" -ne 180 ]i && [ "$corefile" -eq 1 ] ; do
```
madanagopalt: please remove character i after 180 count check

ci/unittests_linux.sh
```diff
@@ -24,16 +26,31 @@ TESTLOGS=$TRAVIS_BUILD_DIR/logs/test_logs;
...
-echo "kill -9 `ps -ef | grep pxscene2dtests |grep -v grep|grep -v pxscene2dtests.sh|awk '{print $2}'`"
-kill -9 `ps -ef | grep pxscene2dtests |grep -v grep|grep -v pxscene2dtests.sh|awk '{print $2}'`
+#check for corefile presence
+processId=`ps -ef | grep pxscene2dtests |grep -v grep|grep -v pxscene2dtests.sh|awk '{print $2}'`
+ls -l /tmp/pxscenecrash
+retVal=$?
+if [ "$retVal" -eq 0 ]
+  then
+  $TRAVIS_BUILD_DIR/ci/check_dump_cores_linux.sh `pwd` "$TRAVIS_BUILD_DIR/tests/pxScene2d/pxscene2dtests" "$processId" "$TESTLOGS"
+  kill -9 $processId
```
madanagopalt: just want to make sure, if we send signal to a process which is already in signal handler, it is not creating any problem?.

ci/check_dump_cores_linux.sh
```diff
@@ -3,31 +3,24 @@
...
-	gdb -q --command="$TRAVIS_BUILD_DIR/ci/debuggercmds_linux" -c $coreFileName $binary  2&>gdblogs
-	cat gdblogs
-	if [ "$TRAVIS_PULL_REQUEST" != "false" ]
+        sudo gdb -q --command="$TRAVIS_BUILD_DIR/ci/debuggercmds_linux" --pid=$3 $2 2&> gdblogs
+        if [ "$TRAVIS_PULL_REQUEST" != "false" ]
 		then
 		echo "**********************PRINTING CORE STACK DETAILS**************************"
-		echo "PATH :  $path"
-		cat $tempLogFile
+		cat gdblogs
```
madanagopalt: please clarify the difference between gdb.txt output and gdblogs output. is it ok, if we don't have gdb.txt?

ci/check_dump_cores_linux.sh
```diff
@@ -3,31 +3,24 @@
...
 if [ "$retVal" -eq 0 ]
 	then
-	tempLogFile="gdb.txt"
-	coreFileName="core"
-	sudo chmod 777 $coreFileName
-	echo "gdb -q --command=\"$TRAVIS_BUILD_DIR/ci/debuggercmds_linux\" -c $coreFileName $binary  2&>gdblogs"
-	gdb -q --command="$TRAVIS_BUILD_DIR/ci/debuggercmds_linux" -c $coreFileName $binary  2&>gdblogs
-	cat gdblogs
-	if [ "$TRAVIS_PULL_REQUEST" != "false" ]
+        sudo gdb -q --command="$TRAVIS_BUILD_DIR/ci/debuggercmds_linux" --pid=$3 $2 2&> gdblogs
```
madanagopalt: can we use variable names here instead of shell args, just for readability :)

ci/unittests_linux.sh
arun-govindan: In this case, if the signal sent is a SIGKILL or SIGTERM, it will kill the process and we will not get the stack trace as this signal cannot be ignored and this script will exit gracefully. 
Whereas with other signals like SIGSEGV, the script is continued to get the stack trace.

ci/check_dump_cores_linux.sh
arun-govindan: This is actually a code cleanup. gdb.txt is stored in a variable tempLogFile, but there is no data redirected to tempLogFile or gdb.txt. The output of gdb is sent to gdblogs. So i removed the unused variables.

PR #1359, opened 2018-08-02
fix XModule async callbacks

npoltorapavlo:


examples/pxScene2d/src/rcvrcore/XModule.js
```diff
@@ -16,10 +16,7 @@ limitations under the License.
 
 */
 
-"use strict";
```
mfiess: Out of curiosity, what is causing the need to remove "use strict"?

mfiess:
I'm going to review the freeResources() changes more thoroughly next week.

examples/pxScene2d/src/rcvrcore/XModule.js
npoltorapavlo: I thought it is strict by default in node. However after experimenting I see it is not, brought this back

examples/pxScene2d/src/rcvrcore/XModule.js
```diff
@@ -28,63 +27,24 @@ function XModule(name, appSceneContext, basePath, jarName) {
...
-    {
-      this.promises.pop();
-    }
-  }
-  this.promises = null;
-  this.moduleData = null;
   this.appSceneContext = null;
-  this.imports = null;
   this.log = null;
-  this.importReplacementMap = null;
```
mfiess: Can you please briefly describe why we no longer want to clean up these references inside of freeResources?

npoltorapavlo: imports, moduleData, promises, moduleNameList, pendingIncludes have been removed. As to 'exports', '= null' is sufficient in javascript.

PR #1370, opened 2018-08-08
Changes for loading resources from jar

kannanmadanagopalt:
just initial changes. it is under testing.


examples/pxScene2d/src/pxResource.h
```diff
@@ -139,6 +139,7 @@ class pxResource : public rtObject
   rtMutex mDownloadInProgressMutex;
   mutable rtMutex mLoadStatusMutex;
   rtCORSRef mCORS;
+  rtObjectRef mArchive;
```
mfiess: Let's remove storing the archive as a member variable inside of pxResource.  If the image or font is going to be fetched from the archive then let's pass it in as a parameter of getImage() etc but not store it as a member variable inside of pxResource.

PR #1371, opened 2018-08-09
Changes for fixing font cut-off on less scale values 

kannanmadanagopalt:


examples/pxScene2d/src/pxText.cpp
```diff
@@ -115,6 +115,34 @@ void pxText::resourceReady(rtString readyResolution)
   }     
 }
 
+void pxText::update(double t)
+{
```
mfiess: Can you please move the FBO creation and rendering from the update() method to draw()?

PR #1397, opened 2018-08-15
Fixes for spark and sparkEdge name conversion issues

arun-govindan:


examples/pxScene2d/src/mkapp.sh
```diff
@@ -4,14 +4,15 @@
 minJS=./jsMin.sh  #minify
 
 externalDir=../external
-appName=Spark
+APPNAME=Spark
 if [ "$TRAVIS_EVENT_TYPE" == "cron" ]
 then
-appName=SparkEdge
+APPNAME=SparkEdge
 else
```
conniefry: Do you really need this 'else' since APPNAME is already "Spark"

PR #1420, opened 2018-08-27
disallow create of optimus app with duplicate ID

npoltorapavlo:


examples/pxScene2d/src/rcvrcore/optimus.js
```diff
@@ -482,6 +482,9 @@ function Optimus() {
    * @returns {Application}
    */
   this.createApplication = function(props){
+    if (this.getApplicationById(props.id) !== null) {
+      throw new Error("An application with ID '" + props.id + "' already exists");
+    }
```
mfiess: Let's not throw an exception here.  Instead, we should still create the app but reject its ready promise.  Also, we should still return the app but not add the app to the applicationsArray.

examples/pxScene2d/src/rcvrcore/optimus.js
```diff
@@ -316,91 +321,108 @@ function Application(props) {
...
+    this.type = ApplicationType.SPARK;
+    _externalApp = scene.create({t:"scene", parent:root, url:uri, serviceContext:serviceContext});
+    _externalApp.on("onReady", function () { _this.log("onReady"); }); // is never called
+    _externalApp.on("onClientStarted", function () { _this.log("onClientStarted"); }); // is never called
+    _externalApp.on("onClientConnected", function () { _this.log("onClientConnected"); }); // is never called
+    _externalApp.on("onClientDisconnected", function () { _this.log("onClientDisconnected"); }); // is never called
+    _externalApp.on("onClientStopped", function () { _this.log( "onClientStopped"); }); // is never called
+    _externalApp.ready.then(function() {
+      _this.log("successfully created Spark app: " + _this.id);
+      _readyResolve();
```
mfiess: Let's check to ensure _readyResolve is defined before calling it (also in other usages).

examples/pxScene2d/src/rcvrcore/optimus.js
```diff
@@ -316,91 +321,108 @@ function Application(props) {
...
+      _this.applicationReady();
+    }, function rejection() {
+      _this.log("failed to launch Spark app: " + _this.id);
+      _readyReject(new Error("failed to create"));
```
mfiess: Let's check to ensure _readyReject is defined before calling it (also in other usages).

npoltorapavlo: It is always defined bcs promise executor is called immediately.

examples/pxScene2d/src/rcvrcore/optimus.js
npoltorapavlo: It is always defined bcs promise executor is called immediately.

PR #1431, opened 2018-09-06
fix in about.js

ashwinikalokhe:
Added fix in about.js


mfiess:
This change is now available as part of c2de7600.

PR #1442, opened 2018-09-07
fix for log message not printing

arun-govindan:


examples/pxScene2d/src/rcvrcore/AppSceneContext.js
```diff
@@ -315,9 +315,13 @@ AppSceneContext.prototype.runScriptInNewVMContext = function (packageUri, module
   try {
     if (!isDuk && !isV8) {
       var requireMethod = function (pkg) {
-        log.message(3, "old use of require not supported: " + pkg);
-        // TODO: remove
-        return requireIt(pkg);
+        if (typeof requireIt === "function") { 
+          // TODO: remove
```
mfiess: Let's log the warning even if requireIt is defined.

mfiess: Let's log the require message even if requireIt is defined.

conniefry: Okay, I agree, since the case when .pxsceneEnableRequire is defined would replace the 'requireMethod' directly with 'require' at line 336

arun-govindan: Fixed review comments.

PR #1449, opened 2018-09-10
Fix for SVG caching bug.

FitzerIRL:
'nanosvg' performs a destructive parse to the SVG source buffer.  

Passing it a working copy prevents the destruction propagating elsewhere (e.g. into the cache)

### DIFF
--- a/src/pxUtil.cpp
+++ b/src/pxUtil.cpp
@@ -1023,12 +1023,21 @@ rtError pxLoadSVGImage(const char* buf, size_t buflen, pxOffscreen& o, int  w /*
     return RT_FAIL;
   }
 
-  NSVGimage *image = nsvgParse( (char *) buf, "px", 96.0f); // 96 dpi (suggested default)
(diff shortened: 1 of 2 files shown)


src/pxUtil.cpp
```diff
@@ -1023,13 +1023,22 @@ rtError pxLoadSVGImage(const char* buf, size_t buflen, pxOffscreen& o, int  w /*
...
+  //        Pass it a copy !
+  //
+  void *buf_copy = malloc(buflen);
+  memcpy(buf_copy, buf, buflen);
+  
+  NSVGimage *image = nsvgParse( (char *) buf_copy, "px", 96.0f); // 96 dpi (suggested default)
   if (image == NULL)
   {
     rtLogError("SVG:  Could not init decode SVG.\n");
     return RT_FAIL;
```
dwrobel: At fist glance it looks like there is a missing `free(buf_copy);` just before `return RT_FAIL;`

PR #1478, opened 2018-09-24
To Avoid using msvcrt.lib

arun-govindan:


CMakeLists.txt
```diff
@@ -14,7 +14,12 @@ option(PREFER_SYSTEM_LIBRARIES "Prefer to use system libraries over bundled." "O
 
 option(DISABLE_DEBUG_MODE "Disable debugging mode." "OFF") # OFF for backward compatibility
 
+if (WIN32)
+option(SUPPORT_DUKTAPE "SUPPORT_DUKTAPE" OFF)
```
dwrobel: `Duktape` is not dead yet, so disabling it means you might expect any compilation regression on `Windows` _(if that will be merged in that form)_.

arun-govindan: Hi Damian,

I had a discussion with Mike and Connie and came to know that it is ok to switch off duktape, since now we are moving towards V8.
So I updated the ticket as per the discussion.

dwrobel: I also liked to disable it, but preferable for all platforms. Although, when I asked about it on the [slack](https://rdkm.slack.com/archives/C7RCDGLQ7/p1537796665000100) then the answer was _(or at least how I understood it)_ that's it  too early to disable it.

PR #1501, opened 2018-10-04
CI flow update (Azure DevOps, semaphoreci, wayland-rdp)

dwrobel:
- Release some more resources on semaphoreci (increases free memory),
   -- Recommend to use latest Docker (NATIVE) v1809.1 platform,

 - Allow to run Wayland applications using software renderer only
   (among others it is nedded for Azure DevOps but it might be helpful
    in other places as well),

 - Generic Azure DevOps Pipeline definition for:
   -- Linux (ubuntu-16.04) - we are using our Docker image
      on top of it, though,
   -- Windows (vs2017-win2016),
   -- TODO: macOS - is not enabled as we don't have ready to be used
      single script file which might execute the whole CI flow.

 - Some fixes for building Spark on Azure:
   -- Allow to compile using VS2107 Enterprise version,
   -- Force to use python2 for compiling node 6,
   -- TODO: NSIS installer is missing (needs to be installed).


dwrobel:
TODO List: 
 - on Azure Windows platform:
   - install NSIS - see logs from a [failed build](https://dev.azure.com/damiansparrow/Spark/_build/results?buildId=34&view=logs). _(A stopgap I'm using now is to define Pipeline variable: `SKIP_INSTALLER==True` and modify [build_win32.bat](https://github.com/dwrobel/pxCore/commit/515c8f42cb5f3ec80248ba9c0a1b088780afff78#diff-f65f7ce6504ff8d9752a5a4e684ffb11R94) accordingly - [successful build log](https://dev.azure.com/damiansparrow/Spark/_build/results?buildId=36&view=logs))_,
   - install 7z
 - all Azure platforms:
   - find a way to use a kind of `ccache` between consecutive builds.

PR #1520, opened 2018-10-16
Fixes for Xcode IDE (DEBUG) builds

FitzerIRL:
Changes only affect Xcode IDE project files and use an .xcconfig file to #define *PX_PLATFORM_MAC_XCODE* for Xcode IDE based builds ONLY !


FitzerIRL:
Changes to project files are XCode IDE builds only.  'cmake' is used for Release builds.

Project files changes are to remove "duk" builds.

*Spark.cpp* change is the most important - and ONLY activated for XCode IDE, and hence Debug builds

PR #1541, opened 2018-10-30
Avoid more stringstream appends during download errors

madanagopalt:


src/rtFileDownloader.cpp
```diff
@@ -906,20 +906,26 @@ bool rtFileDownloader::downloadFromNetwork(rtFileDownloadRequest* downloadReques
...
     {
-        stringstream errorStringStream;
+        rtString errorString("Download error for: ");
+        errorString.append(downloadRequest->fileUrl().cString());
 
-        errorStringStream << "Download error for: " << downloadRequest->fileUrl().cString()
-                << ".  Error code : " << res << ".  Using proxy: ";
+        errorString.append(". Error code : ");
+        // keep stringstream only for errorcode
+        stringstream errorCode("");
```
mfiess: If we're looking to avoid stringstreams, then let's remove this one too.

madanwork: made the changes

PR #1545, opened 2018-11-01
Mouse wheel support for windows 

arun-govindan:
Verified with mouse scroll option and it looks working fine.
Whereas with the windows touchpad option, the scrolling is not as fine as the mouse scroll.


arun-govindan:
This PR needs to be merged after the merge of #1540

PR #1551, opened 2018-11-07
PXSCENE_DIRTY_RECTANGLES_DEFAULT_OFF

ashwinikalokhe:


examples/pxScene2d/src/pxScene2d.cpp
```diff
@@ -3286,6 +3286,20 @@ rtError pxScene2d::setShowDirtyRect(bool v)
...
 
+rtError pxScene2d::enableDirtyRect(bool& v) const
+{
+    v=mEnableDirtyRectangles;
+    return RT_OK;
+}
+
+rtError pxScene2d::setEnableDirtyRect(bool v)
+{
+#ifdef PX_DIRTY_RECTANGLES_DEFAULT_ON
```
mfiess: Let's remove this #ifdef check.  We can always take this action since this variable is only used with dirty rectangles.

ashwinikalokhe:
All clear to merge.

PR #1554, opened 2018-11-08
logging url, loadtim and app type

arun-govindan:


examples/pxScene2d/src/rcvrcore/optimus.js
```diff
@@ -504,7 +504,13 @@ function Optimus() {
    * @returns {Application}
    */
   this.createApplication = function(props){
+    var timestamp = scene.clock();
     var app = new Application(props);
+    app.ready.then(function () {
+      var url = props.launchParams.uri ? props.launchParams.uri : props.launchParams.cmd;
+      var loadTime = scene.clock() - timestamp;
+      console.log("url : "+url+", "+app.type+", "+loadTime.toFixed(2)+"ms");
```
mfiess: Let's move this log message to inside of the Application's applicationReady function.

PR #1613, opened 2018-12-10
tests and fixes for rtHttp* listeners:

npoltorapavlo:
* tests for rtHttpRequest
* tests for rtHttpResponse
* new methods for rtEmit: clearListeners by event name, 'once' listeners
* fix for rtHttpRequest: emit 'error' instead of 'response' in case error set

### DIFF
--- a/src/rtObject.cpp
+++ b/src/rtObject.cpp
@@ -56,13 +56,14 @@ rtError rtEmit::setListener(const char* eventName, rtIFunction* f)
     e.isProp = true;
     e.markForDelete = false;
     e.fnHash = f->hash();
-    mEntries.push_back(e);      
+    e.emitOnce = false;
+    mEntries.push_back(e);
   }
   
   return RT_OK;
 }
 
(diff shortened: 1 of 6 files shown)


src/rtObject.cpp
```diff
@@ -56,13 +56,14 @@ rtError rtEmit::setListener(const char* eventName, rtIFunction* f)
...
-    mEntries.push_back(e);      
+    e.emitOnce = false;
+    mEntries.push_back(e);
   }
   
   return RT_OK;
 }
 
-rtError rtEmit::addListener(const char* eventName, rtIFunction* f)
+rtError rtEmit::addListener(const char* eventName, rtIFunction* f, bool emitOnce)
```
mfiess: In what cases do we only want to emit once?

npoltorapavlo: rtHttp* classes should work like node's EventEmitter which has 'once' as well as 'on'.

PR #1627, opened 2018-12-13
support Buffer for rtHttpRequest::write

npoltorapavlo:


conniefry:
Is there already a tests.json test for this, or does one need to be added?

npoltorapavlo:
> Is there already a tests.json test for this, or does one need to be added?

screencapture_jar test (#1634)

PR #1584, opened 2018-11-27
V8 Demo updates

jmgasper:
A few updates to help with demos that weren't working properly on the V8 build.

PR #1649, opened 2018-12-28
V8 build - Test suite crash fix

jmgasper:
# Crash and fix overview on V8 test suite

From the crash report of V8

```sh
Check failed: Handle not reset in first callback. See comments on |v8::WeakCallbackInfo|..
```

The V8 Handle is not reset. It is caused by if the Promise triggered by one event loop gets resolved by other event loop. It is causing the dependencies between data running between two event loops failed, if one loop didn't complete before other. So, promise not registered by first event loop, before the second event loop sends back the ready event.

The fix is straightforward, use a global flag `gIsPumpingJavaScript` to avoid one loop not completing before other.

Actually, this fix has been applied in the Node version, see `rtScriptNode::pump()` in `pxCore/src/rtScriptV8/rtScriptNode.cppL1064-1105`. But in the V8 version `rtScriptV8::pump()`, it was not applied.


kannanmadanagopalt:
yes, same problem happened with node 👍

npoltorapavlo:
This pull request reverts some old changes that I asked not to merge in PR #1584.
Did you add them on purpose (I.e. it fixes the crash)?

PR #1650, opened 2018-12-28
changes for v8 crash on http request delay

madanagopalt:


src/rtHttpRequest.cpp
```diff
@@ -131,6 +131,7 @@ rtError rtHttpRequest::end()
   req->setAdditionalHttpHeaders(mHeaders);
   req->setMethod(mMethod);
   req->setReadData(mWriteData);
+  AddRef();
```
mfiess: I believe this should be req->AddRef().

src/rtHttpRequest.cpp
```diff
@@ -237,6 +238,7 @@ void rtHttpRequest::onDownloadComplete(rtFileDownloadRequest* downloadRequest)
   } else {
     req->mEmit.send("error", downloadRequest->errorString());
   }
+  req->Release();
```
mfiess: Hiding a rtHttpRequest `Release()` inside of `onDownloadComplete` is dangerous as it decrements the reference without the programmer's knowledge.  All APIs that register for `onDownloadComplete` now need to call `AddRef()` before registering for this callback.  Instead, please add a new callback function to rtHttpRequest called `onDownloadCompleteAndRelease`.  This callback will simply call rtHttpRequest's `onDownloadComplete` and then it will call `Release()` on the rtHttpRequest that is part of the passed in rtFileDownloadRequest's callback data.  The method should look like this:

`
void rtHttpRequest::onDownloadCompleteAndRelease(rtFileDownloadRequest* downloadRequest)
{
  onDownloadComplete(downloadRequest);
  rtHttpRequest* req = (rtHttpRequest*)downloadRequest->callbackData();
  req->Release();
}
`

tests/pxScene2d/test_rtHttpRequest.cpp
```diff
@@ -242,39 +242,45 @@ class rtHttpRequestTest : public testing::Test
     ref = req;
 
     rtFileDownloadRequest* dwnl_OK = new rtFileDownloadRequest("https://example.com", req, rtHttpRequest::onDownloadComplete);
-
+    
+    req->AddRef();
```
mfiess: Please remove all changes to test_rtHttpRequest and follow the approach outlined in the previous comment.  If you follow this approach then no changes should be needed in the unit tests.

PR #1662, opened 2019-01-04
enabling image cache tests dependent on network

madanagopalt:


src/rtHttpCache.h
```diff
@@ -97,7 +97,7 @@ class rtHttpCacheData
     rtError calculateRevalidationNeed(bool&,bool&);
 
     /* initiate and handle download */
-    bool handleDownloadRequest(std::vector<rtString>& headers,bool downloadBody=true);
+    virtual bool handleDownloadRequest(std::vector<rtString>& headers,bool downloadBody=true);
```
mfiess: Can we provide a test without changing the signature of handleDownloadRequest?

kannanmadanagopalt: will try and update

tests/pxScene2d/test_imagecache.cpp
```diff
@@ -55,6 +56,25 @@ extern bool continueDownloadHandleCheck;
 // it doesn't guarantee that the returned memory will be available.
 // #define UNAVAILABLE_MEMORY_TEST_ENABLED
 
+bool ignoreNetworkCall = false;
+typedef CURLcode (*curl_perform_t)(CURL*);
+CURLcode curl_easy_perform(CURL * easy_handle )
```
mfiess: Let's try to do another way instead of override a curl method.  Can we use javascript tests instead of native c++ tests to validate functionality?  If the below tests that are commented out are already covered by testRunner tests then let's remove the commented out code.  However, if they are not covered by testRunner tests then can we create testRunner tests for them instead and then remove the commented out tests?

kannanmadanagopalt:
closing this, as different approach is decided to achieve this coverage via test scripts

PR #1663, opened 2019-01-04
Adding support for #RGB, #RRGGBB, and Web Named colors

FitzerIRL:
Functionality per -CSS Color Module Level 3- support ...

https://developer.mozilla.org/en-US/docs/Web/CSS/color_value


FitzerIRL:
Status ?

src/rtString.h
```diff
@@ -115,6 +116,8 @@ class rtString
 
   rtString substring(size_t pos, size_t len = 0) const;
 
+  void toLower() { char *p = mData; while(*p) { *p = tolower(*p); p++; } };
```
mfiess: toLower() won't return the desired result for UTF-8 strings.  Can you please document or rename method to `toLowerAscii()`?

FitzerIRL: Done

PR #1683, opened 2019-01-15
fix: Allow launching via a symlink

trentontrees:
- Added ability to symlink to spark.sh and launch the symlink bin file
- Added handling of relative files parameters


trentontrees:
I found that the spark binary is able to accept query params attached to the file, which is not supported by my shell script updates.

PR #1684, opened 2019-01-21
bundling support

kannanmadanagopalt:


examples/pxScene2d/src/pxArchive.cpp
```diff
@@ -110,7 +110,17 @@ rtError pxArchive::initFromUrl(const rtString& url, const rtCORSRef& cors, rtObj
   mLoadStatus = new rtMapObject;
 
   mUrl = url;
-
+  std::string tempUrl(url.cString());
```
FitzerIRL: It would be nicer to use **rtString** ... if all the functionality is present

kannanmadanagopalt: api similar to lastIndexOf() in std::string is not there

PR #1685, opened 2019-01-23
fix: Add handling of symlink to spark.sh

trentontrees:
Use `realpath` to resolve any symlinks to get the actual directory of
this shell script.  This allows for symlinks to spark.sh to work
correctly instead of using the directory of the symlink.


trentontrees:
Not sure why, but travis linux build blew up on the first commit which was only suppose to effect OSX.  On the second attempt I'm using a set of tools that should be available for OSX and Linux

PR #1687, opened 2019-01-24
changes for spark log identification

kannanmadanagopalt:


examples/pxScene2d/src/node_modules/console_wrap.js
```diff
@@ -0,0 +1,60 @@
...
+module.exports = {
+    trace: trace,
+    warn: warn,
+    error: error,
+    log: log,
+    info: info,
+    dir: console.dir,
+    assert: console.assert,
+    time: console.time,
+    timeEnd: console.timeEnd
```
mfiess: How about other console api functionality?  It looks like it will now be lost.  It's true that we may not be using it today, but if an app wishes to access other apis in console in the future then it will require a change to the console wrapper.

kannanmadanagopalt: verified and made it in sync with console api from nodejs

PR #1692, opened 2019-01-28
Changes needed for [ animate() ] API

FitzerIRL:


FitzerIRL:
Combined into a single PR

PR #1704, opened 2019-02-06
 XRE2-1791 and updated README.md

sergiygladkyy:
Implements RT_LOG_LEVEL support in shell.js


examples/pxScene2d/src/shell.js
```diff
@@ -26,14 +26,59 @@ px.import({ scene: 'px:scene.1.js',
   var scene = imports.scene;
   var keys  = imports.keys;
 
+  var Logger = function (logLevel) {
```
npoltorapavlo: Why don't you use an existing Logger from rcvrcore?

sergiygladkyy: added another implementation based on Logger.js

PR #1721, opened 2019-02-19
Marking overlapping child objects dirty if parent is dirty

ashwinikalokhe:


mfiess:
This is fixed by PR 1722.

PR #1725, opened 2019-02-20
Fixed an alignment issue with newline text special case

ashwinikalokhe:


examples/pxScene2d/src/pxTextBox.cpp
```diff
@@ -591,7 +593,7 @@ void pxTextBox::measureTextWithWrapOrNewLine(const char *text, float sx, float s
       lastLineNumber = lineNumber;
       if( mTruncation == pxConstantsTruncation::NONE && !mWordWrap ) {
         //rtLogDebug("CLF! Sending tempX instead of this->w(): %f\n", tempX);
-        renderOneLine(accString.cString(), 0, tempY, sx, sy, size, lineWidth, render);
+        renderOneLine(accString.cString(), 0, tempY, sx, sy, size, lineWidth, render, isNewLineDetected);
```
conniefry: isNewLineDetected could be stale from a char that was not necessarily the last char processed.  Is that the intention?

ashwinikalokhe: isNewLineDetected indicates that very previous line process had newline char or not. if so then that gets in account while processing very last line

PR #1726, opened 2019-02-20
Break out [pxObject] from [pxScene2d]. Remove [pxPath]...  Tidy up.

FitzerIRL:
Replaces PR from *temp_pxObjectSplitOut* which I will CLOSE


examples/pxScene2d/src/pxScene2d.h
```diff
@@ -153,6 +155,7 @@ class rtFileDownloadRequest;
 class pxScene2d;
 class pxScriptView;
 class pxFontManager;
+#if 0
```
mfiess: Let's remove this code from pxScene2d.h instead of #if 0.

FitzerIRL: Of course ... thanks !

FitzerIRL:
Need to watch the code coverage and perhaps supplement.

PR #1727, opened 2019-02-21
Fixed TextBox charLastX and CharLastY issue when TextBox.(x!=0and y!=0)

ashwinikalokhe:


ashwinikalokhe:
Connie:isNewLineDetected could be stale from a char that was not necessarily the last char processed. Is that the intention?

AshwiniKalokhe: isNewLineDetected indicates that very previous line process had newline char or not. if so then that gets in account while processing very last line

PR #1729, opened 2019-02-25
adding unittests for jsobjectwrapper

kannanmadanagopalt:


src/rtScriptV8/rtObjectWrapper.cpp
```diff
@@ -535,11 +535,15 @@ rtError jsObjectWrapper::Set(const char* name, const rtValue* value)
     if (idx.IsEmpty())
       err = RT_ERROR_INVALID_ARG;
     else
-      err = Set(idx->Value(), value);
+    {
+      if (!Set(idx->Value(), value))
```
mfiess: Set() returns a rtError in this case (not a bool).

PR #1737, opened 2019-02-28
XRE2-1791

sergiygladkyy:
Use RT_LOG_LEVEL to control mouse and keyboard logging from shell.js

### DIFF
--- a/examples/pxScene2d/src/rcvrcore/Logger.js
+++ b/examples/pxScene2d/src/rcvrcore/Logger.js
@@ -17,7 +17,7 @@ limitations under the License.
 */
 
 /**
- * Created by tcarro004 on 7/18/15, updated by sgladk001c on 2/28/19.
+ * Created by tcarro004 on 7/18/15, updated by sgladk001c on 3/01/19.
  */
 
 var loggingLevel = 1;
--- a/examples/pxScene2d/src/rcvrcore/Logger.js
+++ b/examples/pxScene2d/src/rcvrcore/Logger.js
@@ -17,7 +17,7 @@ limitations under the License.
 */
 
 /**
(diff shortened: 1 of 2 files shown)


examples/pxScene2d/src/rcvrcore/Logger.js
```diff
@@ -53,8 +76,4 @@ Logger.prototype.message = function(levelNum, message) {
   console.log(this.fullMessage('MESSAGE:'+levelNum, message));
```
mfiess: Let's encapsulate some of the details here.  Can you please have `Logger.prototype.message = function(levelNum, message)` take in either a string or a number?  If it's string then let's convert to a number using `rtl`.

examples/pxScene2d/src/shell.js
```diff
@@ -20,24 +20,32 @@ var isDuk=(typeof Duktape != "undefined")?true:false;
...
+  var setLoggingLevel = imports.log.setLoggingLevel;
+
+  var logger = new Logger('shell.js');
+  var rtLogLevel = process.env.RT_LOG_LEVEL ?  process.env.RT_LOG_LEVEL : 'warn';
+  setLoggingLevel(rtl(rtLogLevel));
 
   function uncaughtException(err) {
     if (!isDuk && !isV8) {
-      console.log("Received uncaught exception " + err.stack);
+      logger.message(rtl('error'),"Received uncaught exception " + err.stack);
```
mfiess: Please see the comment above.  Let's remove the developer's need to call `rtl()` here.  The new usage would look like: `logger.message('error',"Received uncaught exception " + err.stack);`.  Please make this change for all other logger.message() updates in this PR as well (that use rtl).

PR #1743, opened 2019-03-05
Mutex lock/unlock is not required during reaching from cache. DELIA-3…

ssureshmsc:
Mutex lock/unlock is not required during reaching from cache. The issue reported in DELIA-33016 got fixed with this fix.


ssureshmsc:
> 
> 
> whether all log lines need to be
> info type?

Yes, we need that for LSA Ad insertion purpose. These logs are added for downloadRequest->deferCacheRead() enabled case, wont affect the other use case.

PR #1765, opened 2019-03-19
Added a check to see if a native app is a valid path before launching

binuinbaraj:


madanagopalt:
please re-trigger the build, above travis failure is fixed a day ago

PR #1768, opened 2019-03-21
XRE2-1555 invalid fontResource causes promise never to fire

sergiygladkyy:


conniefry:
Associated CI test in sparkui.org has been merged. Please add it to tests.json as part of this PR to validate this fix.

PR #1771, opened 2019-03-26
fix for http requests created after sandbox destroy

npoltorapavlo:


npoltorapavlo:
> Hi, can we add the same check before sending request also?

It is added, in fact, here: **var isBlocked = appSceneContext.isTerminated;**

conniefry:
Please add a CI test for this.

PR #1777, opened 2019-03-30
pxWayland/pxWaylandContainer should support return codes for suspend/…

sergiygladkyy:
…resume calls

### DIFF
--- a/examples/pxScene2d/src/pxWayland.cpp
+++ b/examples/pxScene2d/src/pxWayland.cpp
@@ -763,17 +763,21 @@ rtError pxWayland::useDispatchThread(bool use)
   return RT_OK;
 }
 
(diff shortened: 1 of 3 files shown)


examples/pxScene2d/src/pxWayland.cpp
```diff
@@ -790,16 +794,17 @@ rtError pxWayland::setProperty(const rtString &prop, const rtValue &val)
   return errorCode;
 }
 
-rtError pxWayland::callMethod(const char* messageName, int numArgs, const rtValue* args)
+rtError pxWayland::callMethodReturns(const char* messageName, int numArgs, const rtValue* args, rtValue& result)
 {
   rtError errorCode = RT_FAIL;
```
mfiess: Let's default result to false at the top of this method.

PR #1778, opened 2019-04-01
add epsilon for time comparision for unittest

madanagopalt:


tests/pxScene2d/test_pxTimerNative.cpp
```diff
@@ -52,7 +52,8 @@ class pxTimerNativeTest : public testing::Test
         startTime = pxSeconds();
         pxSleepMS(TEST_TIME * 1000);
         endTime = pxSeconds();
-        EXPECT_TRUE((int)(endTime - startTime) == TEST_TIME);
+        bool result = (((int)((endTime - startTime)+0.001) == TEST_TIME)) || (((int)((endTime - startTime)-0.001) == TEST_TIME));
+        EXPECT_TRUE(result == true);
```
mfiess: This change still requires too much accuracy.  Please change to something like

`abs(TEST_TIME - (endTime-startTime)) <= 0.001`

madanagopalt:
Time execution between below two instructions are varying in linux and mac. Below condition is passing for linux and failing for mac.

startTime = pxSeconds();
pxSleepMS(TEST_TIME * 1000);
endTime = pxSeconds();
EXPECT_TRUE(fabs(TEST_TIME - (endTime-startTime)) <= 0.001);

In mac, difference between endtime and starttime is coming as 2.05xxxx or 2.07xxxx sometimes.So,fabs(TEST_TIME - (endTime-startTime)) is coming around 0.05 or 0.07.
In linux, difference is coming as 0.000119 and so, it is getting validated.

PR #1797, opened 2019-04-17
Local storage

npoltorapavlo:


src/rtStorage.cpp
```diff
@@ -0,0 +1,461 @@
...
+    return RT_ERROR_INVALID_ARG;
+
+  uint32_t size = 0;
+  rtError retStat = getCurrentSize(size);
+
+  if (RT_OK == retStat)
+  {
+    rtString keyStr = key;
+    rtString valueStr = value.toString();
+    uint32_t newSize = size + keyStr.byteLength() + valueStr.byteLength();
```
conniefry: This does not seem to account for the size difference if key already exists with a different value.

src/rtStorage.h
```diff
@@ -0,0 +1,78 @@
...
+  rtError getItems(const char* keyPrefix, rtObjectRef& retValue) const;
+
+  rtError removeItem(const char* key);
+
+  // Clear all data
+  rtError clear();
+
+  rtError runVacuumCommand();
+
+  static bool isEncryped(const char* fileName);
```
conniefry: Is isEncryped missing a 't'?

examples/pxScene2d/src/pxScene2d.h
```diff
@@ -1198,6 +1208,10 @@ class pxScene2d: public rtObject, public pxIView, public rtIServiceProvider
     return mArchive;
   }
 
+#ifdef PXSCENE_SUPPORT_STORAGE
+  rtError storage(rtObjectRef& v) const;
```
mfiess: Please ensure that all methods are declared without surrounding #ifdefs.  Please add the #ifdef checks inside the methods themselves.  If the define is not declared then please log a message but perform no action.  However, a method should be declared if the define is set or not.  This will prevent us from having to add new defines when building other software components that use Spark headers.  Please update for all other new methods as well.

examples/pxScene2d/src/pxScene2d.cpp
```diff
@@ -2522,6 +2542,66 @@ rtError pxScene2d::getAvailableApplications(rtString& availableApplications)
   return RT_OK;
 }
 
+#ifdef PXSCENE_SUPPORT_STORAGE
```
mfiess: Please ensure that all methods are declared and defined without surrounding #ifdefs.  Please add the #ifdef checks inside the methods themselves.  If the define is not declared then please log a message but do nothing.  However, a method should be declared if the define is set or not.  This will prevent us from having to add new defines when building other software components that use Spark headers.  Please update for all other new methods as well.

examples/pxScene2d/src/pxScene2d.cpp
```diff
@@ -2577,6 +2657,10 @@ rtDefineProperty(pxScene2d, cors);
 rtDefineMethod(pxScene2d, addServiceProvider);
 rtDefineMethod(pxScene2d, removeServiceProvider);
 
+#ifdef PXSCENE_SUPPORT_STORAGE
+rtDefineProperty(pxScene2d, storage);
```
mfiess: Let's expose this property always but have it return RT_FAIL if storage is compiled off.

examples/pxScene2d/src/pxScene2d.h
```diff
@@ -899,6 +905,10 @@ class pxScene2d: public rtObject, public pxIView, public rtIServiceProvider
   rtMethod1ArgAndNoReturn("addServiceProvider", addServiceProvider, rtFunctionRef);
   rtMethod1ArgAndNoReturn("removeServiceProvider", removeServiceProvider, rtFunctionRef);
 
+#ifdef PXSCENE_SUPPORT_STORAGE
+  rtReadOnlyProperty(storage,storage,rtObjectRef);
```
mfiess: Let's expose this property always but have it return RT_FAIL if storage is compiled off.

PR #1802, opened 2019-04-23
tests for leak from removeAll after gc

madanagopalt:


madanagopalt:
please wait till the changes on sparkui site for test files got merged before merging this

conniefry:
This change will not be backward compatible for apps already using the existing API.

PR #1803, opened 2019-04-24
typecast error fixes

moorthy-bs:
* unrecognized identifier due to gcc 8
* typecast error fix for 64bit


moorthy-bs:
> please make changes for libnode between MODIFIED CODE BEGIN and MODIFIED CODE END tags

Comments added

PR #1804, opened 2019-04-24
changes for migrating landing page from pxscene.org to sparkui.org

madanagopalt:


madanagopalt:
please wait till the changes on sparkui got merged before merging this

PR #1835, opened 2019-05-16
Adding a logging to capture font atlas size/usage	

ashwinikalokhe:


examples/pxScene2d/src/pxFont.cpp
```diff
@@ -883,6 +883,7 @@ bool pxFontAtlas::addGlyph(uint32_t w, uint32_t h, void* buffer, GlyphTextureEnt
   {
     mTexture = context.createTexture(PXSCENE_FONT_ATLAS_DIM,PXSCENE_FONT_ATLAS_DIM,PXSCENE_FONT_ATLAS_DIM,PXSCENE_FONT_ATLAS_DIM, NULL);
   }
+  rtLogDebug("adding a texture of [%d] size memory", w*h);
```
mfiess: w*h will be unsigned but %d expects signed.

ashwinikalokhe: Fixed it with %u

PR #1842, opened 2019-05-22
changes for remote debugging (testing)

tmadanagopal:


src/rtScript.h
```diff
@@ -98,9 +106,21 @@ class rtScript
 
   void* getParameter(rtString param);
 
+#ifdef ENABLE_DEBUG_MODE
```
mfiess: Rather than defining methods or not defining methods depending on build defines, let's instead always define these methods, and let's have the methods themselves log a message if the ability is disabled and take no action.  Please follow the same pattern for all method declarations and definitions.

src/rtScript.h
```diff
@@ -98,9 +106,21 @@ class rtScript
...
+  rtError enableDebugger(bool);
+#endif
+
 private:
+#ifdef ENABLE_DEBUG_MODE
```
mfiess: Rather than defining methods or not defining methods depending on build defines, let's instead always define these methods, and let's have the methods themselves log a message if the ability is disabled and take no action.  Please follow the same pattern for all method declarations and definitions.

src/rtScript.h
```diff
@@ -74,6 +79,9 @@ class rtIScript
   virtual rtError pump() = 0;
 
   virtual rtError collectGarbage() = 0;
+#ifdef ENABLE_DEBUG_MODE
```
mfiess: Rather than defining methods or not defining methods depending on build defines, let's instead always define these methods, and let's have the methods themselves log a message if the ability is disabled and take no action.  Please follow the same pattern for all method declarations and definitions.

PR #1850, opened 2019-05-28
uiReady

comcastmike:


examples/pxScene2d/src/rcvrcore/optimus.js
```diff
@@ -381,12 +446,21 @@ function Application(props) {
     _externalApp.on("onClientStopped", function () { _this.log( "onClientStopped"); }); // is never called
     _externalApp.ready.then(function() {
       _this.log("successfully created Spark app: " + _this.id);
-      _readyResolve();
+      _readyBaseResolve();
```
mfiess: Do we need to add a check here to confirm _readyBaseReject is valid before calling?  Can it still be undefined at this point?

examples/pxScene2d/src/rcvrcore/optimus.js
```diff
@@ -381,12 +446,21 @@ function Application(props) {
...
+          function(result) { _uiReadyResolve(result); }, 
+          function(err) { _uiReadyReject(err); }
+        );
+      }
     }, function rejection() {
       var msg = "Failed to load uri: " + uri;
       _this.log("failed to launch Spark app: " + _this.id + ". " + msg);
-      _readyReject(new Error("failed to create. " + msg));
+      _readyBaseReject(new Error("failed to create. " + msg));
+      _uiReadyReject();
```
mfiess: Do we need to add a check here to confirm _readyBaseReject and _uiReadyReject are valid before calling them?  Can they still be undefined at this point?

examples/pxScene2d/src/rcvrcore/optimus.js
```diff
@@ -408,16 +482,28 @@ function Application(props) {
     _externalApp.ready.then(function() {
     }, function rejection() {
       _this.log("failed to launch Spark instance app: " + _this.id);
-      _readyReject(new Error("failed to create"));
+      _readyBaseReject(new Error("failed to create"));
+      _uiReadyReject();
```
mfiess: Do we need to add a check here to confirm _readyBaseReject and _uiReadyReject are valid before calling them?  Can they still be undefined at this point?

examples/pxScene2d/src/rcvrcore/optimus.js
```diff
@@ -408,16 +482,28 @@ function Application(props) {
...
+      _externalApp.api.on("onApplicationLoaded", function(e)
+      {
+        if(e.success)
+          _urlChangeResolve();
+        else
+          _urlChangeReject();
+      });
+      
+      _readyBaseResolve();
+      _uiReadyResolve();
```
mfiess: Is it possible for these callback methods to be undefined or null at this point?  If so, let's add a check to make sure they're valid before calling the functions.

examples/pxScene2d/src/rcvrcore/optimus.js
```diff
@@ -408,16 +482,28 @@ function Application(props) {
...
       _this.applicationReady();
     }, function rejection() {
       _this.log("failed to create Spark instance");
-      _readyReject(new Error("failed to create"));
+      _readyBaseReject(new Error("failed to create"));
+      _uiReadyReject();
```
mfiess: Same as above, do we need a check here to see if these methods are valid?

mfiess:
This change is included as a part of PR 1865.

PR #1857, opened 2019-05-31
buildWindows fixed

ashwinikalokhe:


mfiess:
The fix for this one will be included in PR 1858.

PR #1891, opened 2019-06-10
gif windows build fixes

ashwinikalokhe:


src/CMakeLists.txt
```diff
@@ -37,7 +37,7 @@ set(BUILD_RTCORE_SHARED_LIBRARY 1)
 
 if (BUILD_WITH_GIF)
 message("Building with GIF support")
-
+include_directories(AFTER ${EXTDIR}/giflib-5.1.9/)
```
mfiess: We should only add this include directory for GIF if PREFER_SYSTEM_LIBRARIES is OFF because the system version may be different than the externals version.

ashwinikalokhe:
I think we can ignore the codecov for this PR as none file is affected other than build files.

PR #1894, opened 2019-06-11
Added fix for boundX with truncate on/off wrap on/Off with longest text

ashwinikalokhe:


conniefry:
Closing this.  We may revisit in future.

PR #1896, opened 2019-06-12
Adding change to support Unhandled nativekeycodes

pavannandyala:
Adding this change to support Unhandled native key codes
so it can be handled in the Application.

Signed-off-by: pnandyala <pavan.nandyala@lnttechservices.com>


mfiess:
Which key codes are not handled that you are looking to add?  We'll want to look at adding them a different way rather than setting the common key code to the native key code.

PR #1909, opened 2019-06-17
Adding [shaderResource]

FitzerIRL:


examples/pxScene2d/src/pxContextGL.cpp
```diff
@@ -27,6 +27,8 @@
 #include "rtScript.h"
 #include "rtSettings.h"
 
+#include "pxShaderResource.h"
```
mfiess: Can you please remove the dependency of pxShaderResource (which depends on pxResource) from pxContext?

FitzerIRL: Done

examples/pxScene2d/src/pxObject.h
```diff
@@ -595,10 +607,23 @@ class pxObject: public rtObject
   bool mClip;
   bool mMask;
   bool mDraw;
+
+  rtObjectRef             mEffectRef;         // Shader Config objects
+  rtObjectRef             mEffectConfigRef;   // rtMap   is Shader Config object
+  rtObjectRef             mEffectArrayRef;    // rtArray of Shader Config objects
+
+  rtObjectRef             mEffects;           // rtArray of Shader Config objects
+  rtShaderResource       *mEffectShaderPtr;   // Shader Effect used by this pxObject
```
mfiess: Should this be a rtShaderResourceRef?  Is there a reason this is stored as a raw pointer?

examples/pxScene2d/src/pxShaderResource.h
```diff
@@ -0,0 +1,185 @@
...
+  Unless required by applicable law or agreed to in writing, software
+  distributed under the License is distributed on an "AS IS" BASIS,
+  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  See the License for the specific language governing permissions and
+  limitations under the License.
+
+  Created by Fitzpatrick, Hugh on 5/10/19.
+*/
+
+#ifndef PXSHADERRESOURCE_HPP
```
mfiess: Please name to `PXSHADERRESOURCE_H` (from `PXSHADERRESOURCE_HPP`) since this is a .h file (and not a .hpp file).

examples/pxScene2d/src/pxShaderResource.cpp
```diff
@@ -0,0 +1,1433 @@
...
+  else
+  {
+    mFragmentSrc.term(); // Dump the source data...
+    mVertexSrc.term();   // Dump the source data...
+
+    setLoadStatus("statusCode",0);
+    // Since this object can be released before we get a async completion
+    // We need to maintain this object's lifetime
+    // TODO review overall flow and organization
+    AddRef();
```
mfiess: Just curious, why do we need to call `AddRef();` twice inside of `loadResourceFromFile()`?  I understand a call to `AddRef()` before calling `gUIThreadQueue->addTask()`, but `AddRef()` is always called on this object at the beginning of this method as well.  Why the need for two calls?

examples/pxScene2d/src/pxShaderResource.cpp
```diff
@@ -0,0 +1,1433 @@
...
+    if (gUIThreadQueue)
+    {
+      gUIThreadQueue->addTask(onDownloadCompleteUI, this, (void *) "resolve");
+    }
+  }
+#endif // 0
```
mfiess: Let's log a warning if we are trying to load a shader resource from an archive since it's not supported at this time.  This will help debug the issue quickly in the future.

examples/pxScene2d/src/pxShaderResource.cpp
```diff
@@ -0,0 +1,1433 @@
...
+
+      return RT_OK;
+    }
+  }
+
+  return RT_FAIL;
+}
+
+// sampler
+/*static*/ rtError rtShaderResource::bindTexture5(const uniformLoc_t &p)
```
mfiess: Let's combine `bindTexture3`, `bindTexture4`, and `bindTexture5` into single method that takes an additional argument for the texture id.  All the code between these three methods are identical besides the calls to `glActiveTexture()` and `glUniform1i()`, and they are great candidates to combine to a single method.  This will allow for easier code management going forward since only one method update will be needed instead of making the same changes to three separate methods.

FitzerIRL: Like it !

examples/pxScene2d/src/CMakeLists.txt
```diff
@@ -83,6 +83,8 @@ set(PXSCENE_DEFINITIONS )
 set(PXSCENE_ADDITIONAL_RESOURCES )
 set(PXSCENE_INSTALLER 0)
 
+set(ANIMATION_ROTATE_XYZ )
```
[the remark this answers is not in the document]
FitzerIRL: Done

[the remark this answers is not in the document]
FitzerIRL: Done

examples/pxScene2d/src/pxShaderUtils.h
```diff
@@ -0,0 +1,194 @@
...
+#define PX_SHADERUTILS_H
+
+#include <map>
+
+class rtValue; //fwd
+
+#include "pxCore.h"
+#include "pxTexture.h"
+
+#include <exception>
```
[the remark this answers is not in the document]
FitzerIRL: Done

examples/pxScene2d/src/pxObject.cpp
```diff
@@ -1586,6 +1624,275 @@ void pxObject::repaintParents()
...
+
+      drawShader(fbo);
+    }
+    // - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
+  }
+  context.setFramebuffer(previousRenderSurface);
+}
+
+
+void pxObject::drawShader(pxContextFramebufferRef &flattenFbo)
```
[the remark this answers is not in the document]
FitzerIRL: Done

examples/pxScene2d/src/pxShaderResource.h
FitzerIRL: Done

examples/pxScene2d/src/pxShaderResource.cpp
FitzerIRL: Done

FitzerIRL: Done

examples/pxScene2d/src/pxShaderResource.cpp
FitzerIRL: Removed

examples/pxScene2d/src/pxObject.h
FitzerIRL: Refactored

examples/pxScene2d/src/pxShaderResource.cpp
FitzerIRL: Done

examples/pxScene2d/src/pxShaderUtils.h
```diff
@@ -0,0 +1,194 @@
...
+
+#define DEFAULT_EJECT_TEXTURE_AGE 5
+
+#ifndef ENABLE_DFB
+  #define PXSCENE_DEFAULT_TEXTURE_MEMORY_LIMIT_IN_BYTES (65 * 1024 * 1024)   // GL
+  #define PXSCENE_DEFAULT_TEXTURE_MEMORY_LIMIT_THRESHOLD_PADDING_IN_BYTES (5 * 1024 * 1024)
+#else
+  #define PXSCENE_DEFAULT_TEXTURE_MEMORY_LIMIT_IN_BYTES (15 * 1024 * 1024)   // DFB .. Should be 40 ?
+  #define PXSCENE_DEFAULT_TEXTURE_MEMORY_LIMIT_THRESHOLD_PADDING_IN_BYTES (5 * 1024 * 1024)
+#endif
```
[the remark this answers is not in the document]
FitzerIRL: Done

src/pxMatrix4T.h
```diff
@@ -18,6 +18,8 @@
 
 // pxMatrix4T.h
 
+#define ANIMATION_ROTATE_XYZ
```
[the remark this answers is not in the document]
FitzerIRL: Done

examples/pxScene2d/src/pxShaderResource.cpp
```diff
@@ -0,0 +1,1389 @@
...
+}
+
+void pxShaderResource::loadResourceFromFile()
+{
+  init();
+
+  // Since this object can be released before we get a async completion
+  // We need to maintain this object's lifetime
+  // TODO review overall flow and organization
+  AddRef();
```
mfiess: There is an extra `AddRef()` call in `loadResourceFromFile()`.  The `AddRef()` calls before the calls to `gUIThreadQueue->addTask()` look good, but it appears that we're calling `AddRef()` one too many times in `pxShaderResource::loadResourceFromFile()`.  This will result in a resource leak when loading shaders from a file.

FitzerIRL: Thanks for spotting this.  Removed this call.

PR #1911, opened 2019-06-18
Gif support for pxBenchmark

ashwinikalokhe:


examples/pxBenchmark/src/CMakeLists.txt
```diff
@@ -57,14 +57,27 @@ set(PXSCENE_DEFINITIONS )
 set(PXSCENE_ADDITIONAL_RESOURCES )
 set(PXSCENE_INSTALLER 0)
 
+if (BUILD_WITH_GIF)
+message("Building with GIF support")
+set(GIF ${EXTDIR}/giflib-5.1.9/)
```
mfiess: Please add a PREFER_SYSTEM_LIBRARIES check to see if we should be using the system version or the version of GIF from externals.

ashwinikalokhe: we have that check in pxCore/src/Cmakelist.txt .. Do you want it here too? We dont have that in pxScene2d/src/CmakeLists.txt.

mfiess: It looks like GIF is used to the set link directories for GIF.  However, this should only be set to the external location (external/giflib-5.1.9) if we're not using system libraries.

ashwinikalokhe: Yes, we always link to external location (external/giflib-5.1.9). in pxCore/src/CmakeLists.txt we specify or link either to systems libraries or externals based on flag...

if (BUILD_WITH_GIF)
message("Building with GIF support")
if(NOT PREFER_SYSTEM_LIBRARIES)
include_directories(AFTER ${EXTDIR}/giflib-5.1.9/)
endif(NOT PREFER_SYSTEM_LIBRARIES)
add_definitions(-DSUPPORT_GIF)

endif (BUILD_WITH_GIF)

PR #1915, opened 2019-06-19
enable gif support when building mac

mfiess:


mfiess:
This change is no longer needed after PR 1951.

PR #1922, opened 2019-06-25
RDK-24363 : Added support for byte range rtFileDownload.

ssureshmsc:
RDK-24363 requires curl download to be byte range. User enable the byte range if required and pass the range string.


mfiess:
Closing this PR for now as we don't want to merge.

PR #1925, opened 2019-06-26
support file: URLs for resources

npoltorapavlo:


examples/pxScene2d/src/pxResource.cpp
```diff
@@ -558,7 +558,7 @@ void pxResource::loadResource(rtObjectRef archive, bool reloading)
     mArchive = arc;
   }
   //rtLogDebug("rtImageResource::loadResource statusCode should be -1; is statusCode=%d\n",mLoadStatus.get<int32_t>("statusCode"));
-  if (mUrl.beginsWith("http:") || mUrl.beginsWith("https:"))
+  if (mUrl.beginsWith("http:") || mUrl.beginsWith("https:") || mUrl.beginsWith("file:"))
```
mfiess: Let's avoid using the rtFileDownloader for file:// urls.  Instead, let's load via loadResourceFromFile();.

PR #1939, opened 2019-07-04
Added support for byte range curl download(RDK-24363).

ssureshmsc:
CDN requires the curl download request to be less range or equal to (7 * 47 * 4096). So added support for multiple call call with chunk request instead of single bigger download.

### DIFF
--- a/src/rtFileDownloader.cpp
+++ b/src/rtFileDownloader.cpp
@@ -197,7 +197,8 @@ rtFileDownloadRequest::rtFileDownloadRequest(const char* imageUrl, void* callbac
     , mDownloadOnly(false)
     , mActualFileSize(0)
     , mIsByteRangeEnabled(false)
-    , mByteRangeIntervals()
+    , mByteRangeIntervals(0)
+    , mCurlRetry(false)
 {
   mAdditionalHttpHeaders.clear();
 #ifdef ENABLE_HTTP_CACHE
@@ -593,6 +594,16 @@ size_t rtFileDownloadRequest::byteRangeIntervals(void)
   return mByteRangeIntervals;
 }
 
+void rtFileDownloadRequest::setCurlRetryEnable(bool bCurlRetry)
+{
+  mCurlRetry = bCurlRetry;
+}
+
+bool rtFileDownloadRequest::isCurlRetryEnabled(void)
+{
+  return mCurlRetry;
+}
+
 void rtFileDownloadRequest::setDownloadOnly(bool downloadOnly)
 {
   mDownloadOnly = downloadOnly;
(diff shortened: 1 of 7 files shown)


src/rtFileDownloader.cpp
```diff
@@ -676,6 +724,412 @@ void rtFileDownloader::clearFileCache()
...
+}
+
+bool rtFileDownloader::downloadByteRangeFromNetwork(rtFileDownloadRequest* downloadRequest)
+{
+    CURL *curl_handle = NULL;
+    CURLcode res = CURLE_OK;
+    char errorBuffer[CURL_ERROR_SIZE];
+    int multipleIntervals = 1;
+    size_t startRange = 0;
+    std::string byteRange("NULL");
```
mfiess: Please remove std::string usage.  Please use rtString instead.

src/rtFileDownloader.cpp
```diff
@@ -676,6 +724,412 @@ void rtFileDownloader::clearFileCache()
...
+      if(res != CURLE_OK)
+        break;
+
+      if(iLoop == 0)
+      {
+        curl_easy_setopt(curl_handle, CURLOPT_HEADERFUNCTION, NULL);
+        curl_easy_setopt(curl_handle, CURLOPT_HEADERDATA, NULL);
+
+        // Parsing total filesize from Content-Range.
+        std::string httpHeaderStr(reinterpret_cast< char const* >((char *)chunk.headerBuffer));
```
mfiess: Please remove std::string usage. Please use rtString instead.

src/rtFileDownloader.cpp
```diff
@@ -676,6 +724,412 @@ void rtFileDownloader::clearFileCache()
...
+        size_t findContentRange = httpHeaderStr.find("Content-Range:");
+        if(findContentRange != std::string::npos)
+        {
+          std::string strContentRange = httpHeaderStr.substr(findContentRange+1);
```
mfiess: Please remove std::string usage. Please use rtString instead.

src/rtFileDownloader.cpp
```diff
@@ -676,6 +724,412 @@ void rtFileDownloader::clearFileCache()
...
+          size_t findRange = strContentRange.find("/");
+
+          if(findRange != std::string::npos)
+          {
+            std::string substrRange = strContentRange.substr(findRange+1);
```
mfiess: Please remove std::string usage. Please use rtString instead.

src/rtFileDownloader.cpp
```diff
@@ -676,6 +724,412 @@ void rtFileDownloader::clearFileCache()
...
+            std::string strRange = substrRange.substr(0, substrRange.find("\n")-1);
```
mfiess: Please remove std::string usage. Please use rtString instead.

src/rtFileDownloader.cpp
ssureshmsc: Changed to rtString and pushed again.

ssureshmsc: Changed to rtString and pushed again.

src/rtFileDownloader.cpp
ssureshmsc: Changed to rtString and pushed again.

src/rtFileDownloader.cpp
ssureshmsc: Changed to rtString and pushed again.

src/rtFileDownloader.cpp
ssureshmsc: Changed to rtString and pushed again.

src/rtFileDownloader.cpp
ssureshmsc: Changed to rtString and pushed again.

ssureshmsc:
Changed to rtString and pushed again.

src/rtFileDownloader.cpp
ssureshmsc: Changed to rtString and pushed again.

PR #1969, opened 2019-07-25
added some missing webGL functions

goruklu:


goruklu:
The above webGL functions are used in Lightning-Demo-App

PR #1983, opened 2019-08-01
Fixed file load issue with url starting with /

ashwinikalokhe:


examples/pxScene2d/src/browser/mime.js
```diff
@@ -82,7 +81,7 @@ limitations under the License.
     {
       url = "about.js";
     }
-    else if ( (url.indexOf(':') == -1) && url.endsWith(".js") )
+    else if ( url.indexOf('/') !== 0 && url.indexOf('/') !== 0 && (url.indexOf(':') == -1) && url.endsWith(".js") )
```
mfiess: It looks like `url.indexOf('/') !== 0` is listed twice.

PR #1984, opened 2019-08-01
Fix for "FILE://" or "file://" prefixed URL's

FitzerIRL:


examples/pxScene2d/src/pxShaderResource.cpp
```diff
@@ -390,18 +399,27 @@ void pxShaderResource::loadResourceFromFile()
       break;
     }
 
-    loadVtxShader = rtLoadFile(mVertexUrl, mVertexSrc);
+    bool isVtxFILE =  (  mVertexUrl.beginsWith("file://") ||   mVertexUrl.beginsWith("file://") );
+    rtString pathVtx = mFragmentUrl;
```
mfiess: This should be `rtString pathVtx = mVertexUrl;`

FitzerIRL: oops

FitzerIRL: Fixed

PR #1989, opened 2019-08-06
AddRef() - Fix for compailation failed case.

FitzerIRL:


examples/pxScene2d/src/pxShaderResource.cpp
```diff
@@ -575,6 +587,104 @@ void pxShaderResource::loadResourceFromArchive(rtObjectRef /*archiveRef*/)
 #endif // 0
 }
 
+void pxShaderResource::processDownloadedResource(rtFileDownloadRequest* fileDownloadRequest)
+{
```
mfiess: `pxShaderResource::processDownloadedResource()` is an exact copy of the base class version.  Let's remove this version to allow for easy code maintenance.

examples/pxScene2d/src/pxShaderResource.cpp
```diff
@@ -637,8 +745,6 @@ void pxShaderResource::loadResource(rtObjectRef archive, bool reloading)
     mDownloadInProgressMutex.unlock();
     AddRef(); //ensure this object is not deleted while downloading
     rtFileDownloader::instance()->addToDownloadQueue(vtxRequest);
-
-    return;
   }
   else if (isFrgDataURL)
```
mfiess: Should this be an `if` instead of a `else if` if there is both a vertex shader and a fragment shader data url?  If so, check the following `else if` and `else` statements as the logic may need to be updated.  If the intention is to not support a vertex shader and a fragment shader data url with the same resource then this code doesn't need to be modified.  However, I just wanted to point this out in case that was not the intention.

examples/pxScene2d/src/pxShaderResource.h
```diff
@@ -85,6 +85,7 @@ class pxShaderResource: public pxResource, public shaderProgram
   bool needsFbo()   { return mNeedsFbo;   }
   int32_t passes()  { return mPasses;     }
 
+  virtual void processDownloadedResource(rtFileDownloadRequest* fileDownloadRequest);
```
mfiess: Please remove (see comment above) and use the base class version since the implementation is identical.

examples/pxScene2d/src/pxShaderResource.cpp
FitzerIRL: Fixed a callback bug in base class.   The removes the need for pxShaderResource::processDownloadedResource() in my version,

examples/pxScene2d/src/pxShaderResource.h
FitzerIRL: Done

examples/pxScene2d/src/pxShaderResource.cpp
FitzerIRL: Reworked. And added additional testing to check permutations.

PR #2013, opened 2019-08-21
 fix issues with tests when running multiple mjs apps

madanagopaltcomcast:


examples/pxScene2d/src/initGL.js
```diff
@@ -650,6 +671,65 @@ async function loadMjs(source, url, context)
     };
   };
 
+  function loadESM(filename) {
```
npoltorapavlo: Please revert. For some reason you removed my changes on bootstrap. Please merge everything from _webgl to your branch.

PR #2028, opened 2019-09-09
Webgl on windows

ashwinikalokhe:


ashwinikalokhe:
It uses this defines in built time but at run time it wasnt getting to the right location.. It was runtime issue

PR #2035, opened 2019-09-16
increase app timeout for webgl apps

kannanmadanagopalt:


kannanmadanagopalt:
Reason for this change is because, for testRunner execution I see leak is happening and also the complete execution of app is not done before the testrunner exits (i.e seen the gl.init() happened after testrunner exited).

mfiess:
1.

PR #2036, opened 2019-09-17
move test_ws.js to testsDesktop.json bcs no IPv6 support

npoltorapavlo:


npoltorapavlo:
Abandoning as someone solved the problem by removing these tests .

PR #2043, opened 2019-09-25
Mutex lock/unlock is not required during reaching from cache.

ssureshmsc:
Mutex lock/unlock is not required during reaching from cache.


kannanmadanagopalt:
Suresh, any reason for this change?

ssureshmsc:
We had this change earlier in Spark 1.9 https://github.com/pxscene/pxCore/pull/1743, but not in Spark 1.10. There was a issue reported in DELIA-33016, got fixed by this fix. We need this change.

PR #2048, opened 2019-09-27
pxTextCanvas: support for CutSx, CutSy, CutEx, CutEy, fontStyle in Lightning++

sergiygladkyy:


examples/pxScene2d/src/pxFont.h
```diff
@@ -327,7 +336,7 @@ class pxFontManager
   
   public: 
     
-    static rtRef<pxFont> getFont(const char* url, const char* proxy = NULL, const rtCORSRef& cors = NULL, rtObjectRef archive = NULL);
+    static rtRef<pxFont> getFont(const char* url, const char* proxy = NULL, const char* fontStyle = NULL, const rtCORSRef& cors = NULL, rtObjectRef archive = NULL);
```
mfiess: Let's add this new argument to the end of the API because other users of this API may already be using the old definition.  Let's add any new arguments to APIs to the end to ensure existing usage doesn't cause an undesired regression.

PR #2067, opened 2019-10-14
build updates for gif

kannanmadanagopalt:


mfiess:
This change will be included as part of the _webgl merge.

PR #2110, opened 2019-11-06
pxVideo multiple instances

mateusz-hobgarski-red:
Added support for multiple pxVideo objects being created and used at the same time.


examples/pxScene2d/src/pxVideo.h
```diff
@@ -152,15 +153,11 @@ class pxVideo: public pxObject
     	int pixel_h;
     };
     YUVBUFFER mYuvBuffer;
-    bool initialized = false;
-    GThread *aampMainLoopThread;
+    bool mInitialized = false;
```
mfiess: mInitialized is a member variable of rtObject (which pxObject and pxVideo inherit from) and has a different meaning then the use of mInitialized here.  The easiest solution is to rename the member variable mInitialized in pxVideo to a more descriptive name to avoid the conflict.

mateusz-hobgarski-red: Thanks, I will do that.

PR #2111, opened 2019-11-06
initGL optimization changes (need app test)

madanagopalt:
fixed sparkHttp issues with node-fetch


mfiess:
These changes were included as part of PR 2112 (along with some other updates to support all app).

PR #2114, opened 2019-11-08
add EndOfStream and ProgressUpdate aamp's events handling in pxVideo

tomasz-kumor-red:


examples/pxScene2d/src/pxVideo.cpp
```diff
@@ -84,6 +84,7 @@ pxVideo::pxVideo(pxScene2d* scene):pxObject(scene)
 #endif //ENABLE_SPARK_VIDEO_PUNCHTHROUGH
 , mAutoPlay(false)
 , mUrl("")
+, mRtEmit(new rtEmit())
```
tomasz-kumor-red: Here is the issue. When I use mRtEmit member that I added to pxVideo class, I am able to receive events in JS. When I try to use base class member mEmit then it doesn't work.  mEmit is defined as public member in pxObject class.

PR #2115, opened 2019-11-09
fix for hangs

madanagopalt:


examples/pxScene2d/src/pxResource.cpp
```diff
@@ -990,7 +990,12 @@ void pxResource::processDownloadedResource(rtFileDownloadRequest* fileDownloadRe
       }
       else if (result == PX_RESOURCE_LOAD_WAIT)
       {
-        setLoadStatus("statusCode", PX_RESOURCE_STATUS_DOWNLOAD_PROCESSING);
+        // possibility being status code got changed
+        rtValue currStatusCode = getLoadStatus("statusCode");
+        if (-1 == currStatusCode.toInt32())
+        {
+          setLoadStatus("statusCode", PX_RESOURCE_STATUS_DOWNLOAD_PROCESSING);
+        }
```
mfiess: i think the solution should be (to prevent race conditions):
```
  mLoadStatusMutex.lock();
  rtValue currentStatusCode;
  mLoadStatus.get("statusCode", currentStatusCode);
  if (currentStatusCode.toInt32() == -1) 
  {
    //only update the status code if the background processing is not complete
    mLoadStatus.set("statusCode", PX_RESOURCE_STATUS_DOWNLOAD_PROCESSING);
  }
  mLoadStatusMutex.unlock();
```

PR #2138, opened 2019-11-21
avoid executing javascript during gc

madanagopalt:


madanagopalt:
raised another PR 2141

PR #2139, opened 2019-11-21
debug shader resource crash (debugging PR)

madanagopalt:


madanagopalt:
Hugh, yes, it is just for debugging purpose and will close after got the information. I want to create a crash purposefully to get a caller flow who is deleting the shader resource during its usage.

PR #2153, opened 2019-12-03
avoid passing experimental vm modules in command line

madanagopalt:


mfiess:
To reduce risk, we'll just pass in this option via the command line.

PR #2165, opened 2019-12-10
pxWayland: correct work with gUIThreadQueue

Gavriliuk:


Gavriliuk:
> It seems that this is a workaround to fix a symptom. To fix the core problem, please add `pxw->AddRef()` before calling `gUIThreadQueue->addTask()` in `pxWayland::clientStatus()`. Also, please make sure to call `pxw->Release()` at the end of `pxWayland::onClientStatus()`. This will ensure that the pxWayland object is not destroyed while waiting for the onClientStatus callback.

The core problem is "new without delete", the change you've offered doesn't fix it

Root cause: pxWayland must not use removeAllTasksForObject() because each event contains statusData pointer which points to a memory allocated with "new" operator and has to be deleted explicitly with "delete"

The current change offers a fix for the "new without delete" issue

PR #2173, opened 2019-12-18
Windows: uvlib issue fixed

ashwinikalokhe:
…nked though duv


madanagopalt:
is it picking uv headers and symbols  from node's one?

ashwinikalokhe:
> is it picking uv headers and symbols from node's one?

dukluv uses dukluv/build/release/uv.lib
libnode uses its own uvlib

both are of different versions so we cant mix them due to some other dependencies

PR #2191, opened 2020-01-09
Added spark updater

ashwinikalokhe:


ashwinikalokhe:
> How will sparkUpdater.sh know that there is a Spark update available? It doesn't look like it checks the plist file in the cloud.

@mfiess  I have updated script to use UPDATE_URL to download software_update.plist and install spark from code_base url

ashwinikalokhe:
Thanks @madanagopalt  for pointing that out as i missed changes in mkapp.sh

PR #2194, opened 2020-01-14
Adding FALLBACK FONT

FitzerIRL:


examples/pxScene2d/src/pxFont.cpp
```diff
@@ -84,7 +86,8 @@ pxFontAtlas gFontAtlas;
 #endif
 
 pxFont::pxFont(rtString fontUrl, uint32_t id, rtString proxyUrl, rtString fontStyle):pxResource(),mFace(NULL),mPixelSize(0), mFontData(0), mFontDataSize(0),
-             mFontMutex(), mFontDataMutex(), mFontDownloadedData(NULL), mFontDownloadedDataSize(0), mFontDataUrl()
+             mFontMutex(), mFontDataMutex(), mFontDownloadedData(NULL), mFontDownloadedDataSize(0),
+            mFontDataUrl(), mFontStyle(), mFallbackGlyphsCount(0), mFallbackFont(NULL)
```
mfiess: The order of mFallbackGlyphsCount and mFallbackFont should be reversed here to avoid a compiler warning.

FitzerIRL: Done

PR #2196, opened 2020-01-17
framework caching updates

madanagopaltcomcast:


npoltorapavlo:
@madanagopaltcomcast , your initial code uses caching like this: ‘src/lightning-web.js’ -> [many] versions. This PR does: URL -> [many] versions, which makes no sense because files from different URLs of the same version won't be shared. Caching from different URLs was the point of frameworks. My suggestion was to make **frameWorkCache** similar to **modmap**: URL -> [one] version, and copy the 2 blocks from _getModule_ marked as _If the module has been loaded already, return it._ and _Same file from different domains is shared if has 'version'._

madanagopaltcomcast:
Thanks Nikita. is it like you meant, same framework code may be present in two servers http://server1/<code> and  http://server2/<code> with same hash?

madanagopaltcomcast:
Scenario of multiple servers will be handled in current code, with same framework deployed in two different servers with same name property.

npoltorapavlo:
Please do not merge this.
It breaks the current caching , and doesn't fix issues pointed out in ticket.

PR #2226, opened 2020-02-28
openssl issue on libnode 6.9

moorthy-bs:
openssl include directories from 'deps' are
missed in node.gyp. This causes to pick openssl
headers from sysroot (which is openssl 1.1 headers)

Signed-off-by: Moorthy Baskar <moorthy.baskaravenkatraman-sambamoorthy@linaro.org>


mfiess:
This change is no longer needed since we are now using node 10.15.

PR #2248, opened 2020-03-25
Delay suspends on some WebApp suspends v3

comcastmike:


examples/pxScene2d/src/rcvrcore/optimus.js
```diff
@@ -998,4 +1056,72 @@ function Optimus() {
...
+    {
+      console.log("OPTIMUS_HTML5_DELAY_SUSPEND_FILE undefined. Not loading HTML5 Delay Suspend Whitelist.");
+      return;
+    }
+
+    //set / read whitelist
+    try
+    {
+      html5_suspend_whitelist = require(process.env.OPTIMUS_HTML5_DELAY_SUSPEND_FILE);
+    }
```
mfiess: Let's read in a file instead of using `require`.  Also, we should have this file be in JSON format.

PR #2263, opened 2020-04-11
inspector updates

mfiess:


mfiess:
We need to update the node in the Spark-Externals build first before this build will succeed.

madanagopalt:
Crash with this travis build will be fixed by PR https://github.com/pxscene/pxCore/pull/2273

PR #2266, opened 2020-04-17
show complete error stack on exception

madanagopalt:


FitzerIRL:
Can you modify the error messages to use ...

`
  console.log('EXCEPTION: Stack:' + e.stack||e);
  console.log("EXCEPTION: Exception during draw in <<NAME OF FUNCTION>> !!");
`

... for example for *setTimeout* >>

`
          console.log('EXCEPTION: Stack:' + e.stack||e);
          console.log("EXCEPTION: Exception during draw in setTimeout !!");
`


1. setInterval
2. setTimeout
3. setImmediate

